PR #40, opened 2019-09-20
Syntax highlighting of decompiled code using r2 color palettes

### LINKED ISSUE #37: Syntax highlighting of decompiled code in r2 shell
Currently the decompiled code is highlighted only in Cutter and not in the r2 shell. This is exhausting for the eyes when reversing for a couple of hours.

A good solution is to reuse the colours from the user's r2 colour scheme. Respectively, when a user changes their r2 shell colour scheme, the decompiler should apply the new colours to its output. There should be no style difference between the decompiler's and disassembler's outputs. 

I am working on an implementation for this. Suggestions for different ideas are welcome.

v45k0:
### Current state of this PR
- Adds syntax highlighting to `pdg`'s and `pdgo`'s outputs. The current implementation takes advantage of ghidra's XML representation of the decompiled code's AST. 
- Extends `RAnnotatedCode` to include syntax highlighting annotations
- Adds an option to enable/disable syntax highlighting
- Closes #37 

------

### Still need to be done:
- [x] add syntax highlighting to `pdgo`'s output
- [x] make highlighting optional (i.e. disable or enable it globally using a config variable)
- [x] API improvements (see discussion in code)
- [ ] Use a dedicated colour parameter for highlighting keywords (right now it uses the `ec pop` key)
------

### Example

Example of how everything looks like in r2-shell (here I am using a modified version of [r2's base16 material theme](https://github.com/jtalowell/base16-radare2)):

2019-09-20-205803_938x483_scrot


src/AnnotatedCode.c
```diff
@@ -64,6 +64,34 @@ R_API void r_annotated_code_print_json(RAnnotatedCode *code)
 			pj_ks (pj, "type", "offset");
 			pj_kn (pj, "offset", annotation->offset.offset);
 			break;
+        case R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT:
+            pj_ks (pj, "type", "syntax_highlight");
+            switch (annotation->syntax_highlight.type) {
+                case R_SYNTAX_HIGHLIGHT_TYPE_KEYWORD:
+                    pj_ks (pj, "syntax_highlight", "keyword");
```
ITAYC0HEN: instead of re-writing and calling "pj_ks" and "syntax_highlighting", I think it would be better that the switch case will change only the 3rd argument of pj_ks (comment, datatype,...), and then use a single call to pj_ks.

src/AnnotatedCode.c
```diff
@@ -73,3 +101,63 @@ R_API void r_annotated_code_print_json(RAnnotatedCode *code)
...
+    if (code->annotations.len == 0) {
+        r_cons_printf("%s\n", code->code);
+        return;
+    }
+
+
+    size_t cur = 0;
+    size_t len = strlen(code->code);
+
+	RCodeAnnotation *annotation;
```
ITAYC0HEN: not sure if it's only my mobile, but seems like an indention issue?

src/AnnotatedCode.c
```diff
@@ -73,3 +101,63 @@ R_API void r_annotated_code_print_json(RAnnotatedCode *code)
...
+	r_vector_foreach (&code->annotations, annotation) {
+        if (annotation->type != R_CODE_ANNOTATION_TYPE_SYNTAX_HIGHLIGHT)
```
ITAYC0HEN: it's a good practice always to use curly braces, even with a single line condition/loop
https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

src/AnnotatedCode.c
```diff
@@ -73,3 +101,63 @@ R_API void r_annotated_code_print_json(RAnnotatedCode *code)
...
+            break;
+        case R_SYNTAX_HIGHLIGHT_TYPE_CONSTANT_VARIABLE:
+            color = PALETE(num): Color_YELLOW;
+        default:
+            break;
+        }
+
+        // (2/3)
+        // the chunk before the syntax highlighting annotation should not be colored
+        for (; cur < annotation->start && cur < len; cur++)
```
ITAYC0HEN: it's a good practice always to use curly braces, even with a single line condition/loop
https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

src/AnnotatedCode.h
```diff
@@ -24,6 +37,11 @@ typedef struct r_code_annotation_t {
 		{
 			ut64 offset;
 		} offset;
+
+    struct
+    {
+      RSyntaxHighlightType type;
```
ITAYC0HEN: indention is wrong here, right?

src/AnnotatedCode.c
thestr4ng3r: No it's not, I explicitly allow it for r2ghidra ;-P

thestr4ng3r: Sorry nevermind, the AnnotatedCode.c file is meant to be merged into r2 at some point, so there should be no single-line ifs here. But in the rest of r2ghidra it is allowed.

src/AnnotatedCode.h
```diff
@@ -2,15 +2,28 @@
...
+typedef enum r_syntax_highlight_type_t {
+  R_SYNTAX_HIGHLIGHT_TYPE_KEYWORD,
+  R_SYNTAX_HIGHLIGHT_TYPE_COMMENT,
+  R_SYNTAX_HIGHLIGHT_TYPE_DATATYPE,
+  R_SYNTAX_HIGHLIGHT_TYPE_FUNCTION_NAME,
+  R_SYNTAX_HIGHLIGHT_TYPE_FUNCTION_PARAMETER,
+  R_SYNTAX_HIGHLIGHT_TYPE_LOCAL_VARIABLE,
+  R_SYNTAX_HIGHLIGHT_TYPE_CONSTANT_VARIABLE,
+  R_SYNTAX_HIGHLIGHT_TYPE_GLOBAL_VARIABLE,
+} RSyntaxHighlightType;
```
thestr4ng3r: I think it might be better to have all these as separate entries in RCodeAnnotationType and without SYNTAX_HIGHLIGHTING in the name. This way we can also add more info, for example for a variable a reference to the actual value which we can use for example for right-click in Cutter.
Or do you think this fits ghidra's xml better?

src/AnnotatedCode.h
```diff
@@ -37,6 +55,7 @@ R_API void r_annotated_code_free(RAnnotatedCode *code);
 R_API void r_annotated_code_add_annotation(RAnnotatedCode *code, RCodeAnnotation *annotation);
 R_API RPVector *r_annotated_code_annotations_in(RAnnotatedCode *code, size_t offset);
 R_API void r_annotated_code_print_json(RAnnotatedCode *code);
+R_API void r_annotated_code_print_with_syntax_highlighting(RAnnotatedCode *code, RCore *core);
```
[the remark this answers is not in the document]
thestr4ng3r: RCons has a global instance. You can always access it with `r_cons_singleton()`

src/AnnotatedCode.h
thestr4ng3r: Looking at the XML, I think your current approach is good. But I'd still like to hear your opinion.

v45k0: The current implementation fits 1:1 to ghidra's XML data model. However, I don't think it is suitable for anything other than syntax highlighting.

The main problem is that the granularity is too coarse. For example, both string literals and integer constants are classified as `R_SYNTAX_HIGHLIGHT_TYPE_CONSTANT_VARIABLE` by ghidra. So if we want to treat strings and integers in a different manner (e.g. in Cutter with a click or to highlight them in different colours), this will need additional parsing. There are other examples where finer granularity might also be needed (e.g. differentiate between imported vs not-imported symbols).

What you purpose is a good idea. The question is whether we want to stay constrained by ghidra's XML output or we should simply lex the decompiled code and annotate it ourselves. This is more powerful and gives is enough flexibility for future extensions (e.g. we can more easily make the decompiled code to be interactive).

v45k0: For now, I think we can leave the current implementation as it is to provide at least syntax highlighting. Later we can start a separate discussion on how to add a more powerful type annotation for `RAnnotatedCode` (and what we can use it for).

radare:
Why not make the colorizer in r2 so it can be used for other purposes? spp can be used as tokenizer for example

src/AnnotatedCode.h
thestr4ng3r: Yes agreed. One possibility would also be to have both "actual" semantic annotations and syntax-highlighting annotations in parallel. This can make code that only needs syntax highlighting easier to write and semantic annotations can always easily be converted to syntax highlighting ones if needed.

thestr4ng3r:
> Why not make the colorizer in r2 so it can be used for other purposes? spp can be used as tokenizer for example

As said, the plan is to move RAnnotatedCode to r2 later, so it can be used for anything including generating it using a lexer from plain C code.

src/AnnotatedCode.c
```diff
@@ -73,3 +101,63 @@ R_API void r_annotated_code_print_json(RAnnotatedCode *code)
...
+        // now we have a syntax highlighting annotation.
+        // pick a suitable color for it.
+#define PALETE(x) (core->cons && core->cons->context->pal.x)? core->cons->context->pal.x 
+        const char* color = Color_RESET;
+        switch (annotation->syntax_highlight.type) {
+        case R_SYNTAX_HIGHLIGHT_TYPE_COMMENT:
+            color = PALETE(comment): Color_WHITE;
+            break;
+        case R_SYNTAX_HIGHLIGHT_TYPE_KEYWORD:
+            color = PALETE(pop): Color_MAGENTA;
```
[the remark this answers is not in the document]
thestr4ng3r: Hmm yeah I can't find any fitting entry too. I'd prefer to avoid adding a new entry if possible because all themes have to be adapted too, but if necessary it can be done. @radare what do you think?

radare:
About what

thestr4ng3r:
about which palette entry to use or add a new one

v45k0:
This should be ready for merge now @thestr4ng3r @ITAYC0HEN

Issue #39, opened 2019-09-19
No SuperH(4) arch present?


brainstorm:
Thanks for the good work here!

SuperH4 It was introduced recently (11th Sept)... is it too tricky to refresh the build to include it?

```
[0x08000040]> pdgs
DATA:LE:64:default
DATA:BE:64:default
x86:LE:32:default
x86:LE:32:System Management Mode
x86:LE:16:Real Mode
x86:LE:64:default
PIC-16:LE:16:PIC-16
PIC-16:LE:16:PIC-16F
AARCH64:LE:64:v8A
AARCH64:BE:64:v8A
PIC-17:LE:16:PIC-17C7xx
8085:LE:16:default
Dalvik:LE:32:default
6502:LE:16:default
6502:BE:16:default
MIPS:BE:32:default
MIPS:LE:32:default
MIPS:BE:32:R6
MIPS:LE:32:R6
MIPS:BE:64:default
MIPS:LE:64:default
MIPS:BE:64:micro
MIPS:LE:64:micro
MIPS:BE:64:R6
MIPS:LE:64:R6
MIPS:BE:64:64-32addr
MIPS:LE:64:64-32addr
MIPS:LE:64:micro64-32addr
MIPS:BE:64:micro64-32addr
MIPS:BE:64:64-32R6addr
MIPS:LE:64:64-32R6addr
MIPS:BE:32:micro
MIPS:LE:32:micro
z80:LE:16:default
z8401x:LE:16:default
z180:LE:16:default
z182:LE:16:default
68000:BE:32:default
68000:BE:32:MC68030
68000:BE:32:MC68020
68000:BE:32:Coldfire
PIC-24E:LE:24:default
PIC-24F:LE:24:default
PIC-24H:LE:24:default
dsPIC30F:LE:24:default
dsPIC33F:LE:24:default
dsPIC33E:LE:24:default
CR16C:LE:16:default
Toy:BE:32:default
Toy:BE:32:posStack
Toy:LE:32:default
Toy:BE:32:wordSize2
Toy:LE:32:wordSize2
Toy:BE:64:default
Toy:LE:64:default
Toy:BE:32:builder
Toy:LE:32:builder
Toy:BE:32:builder.align2
Toy:LE:32:builder.align2
TI_MSP430:LE:16:default
TI_MSP430X:LE:32:default
pa-risc:BE:32:default
sparc:BE:32:default
sparc:BE:64:default
[...]

PR #41, opened 2019-09-21
Initial SuperH/SuperH4 support. Fixes issue #39.

brainstorm:


brainstorm:
Hmm, couple of warnings still:

```
[0x08000040]> pdg

// WARNING: Type propagation algorithm not settling
// WARNING: Globals starting with '_' overlap smaller symbols at the same address
// WARNING: [r2ghidra] Failed to match radare2 calling convention (null) to Decompiler ProtoModel
// WARNING: [r2ghidra] Failed to find return address in ProtoModel
```

And it's not clear to me how to differentiate between different SuperH's... `flavor` parm I guess, I'll try that.

brainstorm:
There's a few SuperH's variants, I'm not sure how to capture well the different variants with the existing r2 mappings :/

Skärmavbild 2019-09-21 kl  19 40 42

src/R2Architecture.cpp
```diff
@@ -31,6 +31,8 @@ static const std::map<std::string, std::string> arch_map = {
 		{ "ppc64", "PowerPC" } ,
 		{ "sparc32", "sparc" } ,
 		{ "sparc64", "sparc" } ,
+		{ "sh32", "SuperH4" } ,
+		{ "sh32", "SuperH" } ,
```
thestr4ng3r: This cannot work, `std::map` can only have one value per key. For now I would say just use one of these and the other will need to be manually selected.

src/R2Architecture.cpp
```diff
@@ -89,6 +92,11 @@ std::string SleighIdFromCore(RCore *core)
 		be = true;
 	if (!arch_it->second.compare("AARCH64"))
 		flavor = string("v8A");
+	// XXX: Since r2 only has a single sh mapping, this can get tricky with the current structure
+	if (!arch_it->second.compare("SuperH4"))
+		bits = 32; // XXX: SuperH4:LE:32
```
thestr4ng3r: `sh` in r2 only allows 32 bits so I don't think this is necessary. Also you already know that it's 32 bit because only `sh32` maps to `SuperH4`.

thestr4ng3r:
> There's a few SuperH's variants, I'm not sure how to capture well the different variants with the existing r2 mappings :/

Yeah that is why the mapping of ids needs to be refactored. Normally these variants would be covered by asm.cpu.

thestr4ng3r:
Thanks

PR #47, opened 2019-09-30
Reorder code

lironzua:
To make sure `code` isn't being null-referenced


src/core_ghidra.cpp
```diff
@@ -220,9 +220,9 @@ static void Decompile(RCore *core, DecompileMode mode)
 				r_vector_push(r2offsets, &offset);
 			}
 			RAnnotatedCode *code = ParseCodeXML(func, out_stream.str().c_str());
-			code->color_enabled = cfg_var_highlight.GetBool(core->config);
-			if (! code)
+			if (nullptr == code)
```
thestr4ng3r:
```suggestion
			if (!code)
```
please keep it like this

PR #79, opened 2020-02-07
Allow user to use system pugixml library

ret2libc:
**Detailed description**
I want to package this for Fedora and distributions usually want to use as much as possible system libraries. As pugixml is available in Fedora, I want to be able to use that and this patch makes it possible by adding a new cmake option `USE_SYSTEM_PUGIXML`.

**Test plan**
Use `cmake -DUSE_SYSTEM_PUGIXML=ON ..` with pugixml-devel installed (or similar) and install the plugin. It should work as usual.

### DIFF
--- a/third-party/CMakeLists.txt
+++ b/third-party/CMakeLists.txt
@@ -1,8 +1,12 @@
+option(USE_SYSTEM_PUGIXML "Use system pugixml library" OFF)
 
-add_library(pugixml STATIC
-		pugixml/src/pugiconfig.hpp
-		pugixml/src/pugixml.cpp
-		pugixml/src/pugixml.hpp)
-target_include_directories(pugixml PUBLIC pugixml/src)
-set_target_properties(pugixml PROPERTIES POSITION_INDEPENDENT_CODE ON)
-
+if (USE_SYSTEM_PUGIXML)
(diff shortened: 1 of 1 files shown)


third-party/CMakeLists.txt
```diff
@@ -1,8 +1,13 @@
...
-add_library(pugixml STATIC
-		pugixml/src/pugiconfig.hpp
-		pugixml/src/pugixml.cpp
-		pugixml/src/pugixml.hpp)
-target_include_directories(pugixml PUBLIC pugixml/src)
-set_target_properties(pugixml PROPERTIES POSITION_INDEPENDENT_CODE ON)
-
+if (USE_SYSTEM_PUGIXML)
+	find_package(pugixml REQUIRED)
+	include_directories(${PUGIXML_INCLUDE_DIRS})
```
thestr4ng3r: Is this somehow necessary for you? The `pugixml` target exported by the `pugixml` package should already contain the include dir. Also, a `include_directories()` in this file does not affect the parent dirs at all, so it's probably a nop anyway.

ret2libc: I guess you are right ;) I'm not very familiar with cmake. I have pushed an updated version without this line.

PR #101, opened 2020-05-03
Use Debian instead of Ubuntu for Travis and build the Debian package

radare:
**Detailed description**

We need to build things on Debian distros, not ubuntu ones for portability reasons. i pick this one because its the same we are using in radare

But also we want to have the .deb files uploaded and not having to depend on the version of r2.

**Test plan**

Ideally it should use Gitlab CI

**Closing issues**
none

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ branches:
     - master
 
 os: linux
-dist: bionic
+dist: trusty
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ branches:
     - master
 
 os: linux
-dist: bionic
+dist: trusty
```
thestr4ng3r, line 9: trusty is not debian

radare:
Stretch i guess. What a shitty names

ITAYC0HEN (approved):
Looks good :)
I don't understand the reasons behind it, but sure they are legit.

radare:
To get the upload working will require to configure the dist keys in gh settings.

The reason to use debian is for consistency and portability. Ubuntu is a superset of debian. Ubuntu builds dont need to run on debian. All deb files are compiled on debian distros

radare:
Ideally we should use github actions instead of travis here.

thestr4ng3r:
Still broken though

thestr4ng3r:
If you still want to do this, you can add another github workflow similar to the jessie one and just run it in whatever docker base image you want.

PR #108, opened 2020-05-23
Coverity Scan draft

XVilka:
I don't know exact details how r2ghidra-dec is being built, but I put coverity bits here, based on the Cutter file, feel free to fix the building process.

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -26,17 +26,17 @@ jobs:
 #        run: |
 #          chmod go-w $HOME
 #          sudo chmod -R go-w /usr/share
-      
+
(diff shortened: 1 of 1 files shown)


.github/workflows/coverity-scan.yml
```diff
@@ -0,0 +1,65 @@
+name: coverity-scan
+on:
+  schedule:
+    - cron: '0 9 * * 1,5' # Bi-weekly at 09:00 UTC on Monday and Thursday
```
thestr4ng3r, line 4: https://www.youtube.com/watch?v=txa5ia2w7-k

ITAYC0HEN: love it

PR #112, opened 2020-05-30
Emit RAnnotatedCode to Cutter

NirmalManoj:
**Detailed description**

In this PR, RAnnotatedCode is emitted from the decompileAt() function called by Cutter. Made changes to make this possible. 
There is code duplication in `r2ghidra_decompile_annotated_code()` and most of it is from `Decompile()`. I need suggestions in the ways I should refactor both functions to make it more elegant, if possible.
...

**Test plan**

1. Get PR #[2227](https://github.com/radareorg/cutter/pull/2227) and the latest radare2 version before compiling this PR.
2. Open the decompiler widget and make sure that the correct decompiled code is being shown while you select offsets from functions and other widgets.
3. Make sure that the decompiler widget is seeking the correct address while clicking on functions(and other offsets) in the decompiler widget.
4. Make sure that invalid offsets are not causing Segmentation Fault.
5. Check the code and make sure that I have not missed any exceptional cases that will cause errors.

...

**Closing issues**

**Notes**
Merge all these three PRs together to ensure smooth functioning of all decompilers.
Refactored decompiler widget and R2Dec plugin: [Cutter #2227 ](https://github.com/radareorg/cutter/pull/2227)
Refactored R2Ghidra plugin that emits RAnnotatedCode: [R2Ghidra-dec #112](https://github.com/radareorg/r2ghidra-dec/pull/112)
Refactored retdec plugin that emits RAnnotatedCode:[ Florian's fork #1](https://github.com/thestr4ng3r/retdec-r2plugin/pull/1), after this PR is verified, I will send it to retdec's repo.
...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,8 @@ set(CORE_SOURCE
 		src/R2PrintC.h
 		src/R2PrintC.cpp
 		src/RCoreMutex.h
-		src/RCoreMutex.cpp)
+		src/RCoreMutex.cpp
+		src/RAnnotatedCode.h)
 
 if(BUILD_SLEIGH_PLUGIN)
 set(ASM_SOURCE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ set(CORE_SOURCE
 		src/R2PrintC.cpp
 		src/RCoreMutex.h
 		src/RCoreMutex.cpp
-		src/r2ghidra_annotated_code.h)
+		src/r2ghidra.h)
 
 if(BUILD_SLEIGH_PLUGIN)
 set(ASM_SOURCE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ set(CORE_SOURCE
 		src/R2PrintC.cpp
 		src/RCoreMutex.h
 		src/RCoreMutex.cpp
-		src/RAnnotatedCode.h)
+		src/r2ghidra_annotated_code.h)
 
 if(BUILD_SLEIGH_PLUGIN)
 set(ASM_SOURCE
--- a/cutter-plugin/R2GhidraDecompiler.cpp
+++ b/cutter-plugin/R2GhidraDecompiler.cpp
@@ -1,6 +1,7 @@
 /* radare - LGPL - Copyright 2019 - thestr4ng3r */
 
 #include "R2GhidraDecompiler.h"
+#include "../src/RAnnotatedCode.h"
 
 #include <Cutter.h>
 
@@ -16,51 +17,6 @@ R2GhidraDecompiler::R2GhidraDecompiler(QObject *parent)
 
 void R2GhidraDecompiler::decompileAt(RVA addr)
 {
-	if(task)
-		return;
-
-	AnnotatedCode code = {};
-
-	task = new R2Task ("pdgj @ " + QString::number(addr));
-
-	connect(task, &R2Task::finished, this, [this]() {
-		AnnotatedCode code = {};
-		QString s;
-
-		QJsonObject json = task->getResultJson().object();
-		delete task;
-		task = nullptr;
-		if(json.isEmpty())
-		{
-			code.code = tr("Failed to parse JSON from r2ghidra");
-			emit finished(code);
-			return;
-		}
-
-		auto root = json;
-		code.code = root["code"].toString();
-
-		for(QJsonValueRef annotationValue : root["annotations"].toArray())
-		{
-			QJsonObject annotationObject = annotationValue.toObject();
-			CodeAnnotation annotation = {};
-			annotation.start = (size_t)annotationObject["start"].toVariant().toULongLong();
-			annotation.end = (size_t)annotationObject["end"].toVariant().toULongLong();
-			if(annotationObject["type"].toString() == "offset")
-			{
-				annotation.type = CodeAnnotation::Type::Offset;
-				annotation.offset.offset = annotationObject["offset"].toVariant().toULongLong();
-			}
-			else
-				continue;
-			code.annotations.push_back(annotation);
-		}
-
-		for(QJsonValueRef error : json["errors"].toArray())
-			code.code += "// " + error.toString() + "\n";
-
-		emit finished(code);
-	});
-	task->startTask();
-
+	RAnnotatedCode *code = DecompileToRAnnotatedCode(Core()->core());
+	emit finished(code); //Here, we emit RAnnotatedCode *code or by value
 }
--- a/cutter-plugin/R2GhidraDecompiler.cpp
+++ b/cutter-plugin/R2GhidraDecompiler.cpp
@@ -1,7 +1,7 @@
 /* radare - LGPL - Copyright 2019 - thestr4ng3r */
 
 #include "R2GhidraDecompiler.h"
-#include "../src/r2ghidra_annotated_code.h"
+#include "../src/r2ghidra.h"
 
 #include <Cutter.h>
 
--- a/cutter-plugin/R2GhidraDecompiler.cpp
+++ b/cutter-plugin/R2GhidraDecompiler.cpp
@@ -1,7 +1,7 @@
 /* radare - LGPL - Copyright 2019 - thestr4ng3r */
 
 #include "R2GhidraDecompiler.h"
-#include "../src/RAnnotatedCode.h"
+#include "../src/r2ghidra_annotated_code.h"
 
 #include <Cutter.h>
 
@@ -17,6 +17,6 @@ R2GhidraDecompiler::R2GhidraDecompiler(QObject *parent)
 
 void R2GhidraDecompiler::decompileAt(RVA addr)
 {
-	RAnnotatedCode *code = DecompileToRAnnotatedCode(Core()->core());
+	RAnnotatedCode *code = r2ghidra_decompile_annotated_code(Core()->core());
 	emit finished(code); //Here, we emit RAnnotatedCode *code or by value
 }
--- a/cutter-plugin/R2GhidraDecompiler.cpp
+++ b/cutter-plugin/R2GhidraDecompiler.cpp
@@ -15,8 +15,8 @@ R2GhidraDecompiler::R2GhidraDecompiler(QObject *parent)
 	task = nullptr;
 }
 
-void R2GhidraDecompiler::decompileAt(RVA addr)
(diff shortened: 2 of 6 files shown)


cutter-plugin/R2GhidraDecompiler.cpp
```diff
@@ -16,51 +17,6 @@ R2GhidraDecompiler::R2GhidraDecompiler(QObject *parent)
 
 void R2GhidraDecompiler::decompileAt(RVA addr)
```
karliss, line 18: `addr` is ignored.

src/core_ghidra.cpp
```diff
@@ -132,6 +133,87 @@ static void ApplyPrintCConfig(RConfig *cfg, PrintC *print_c)
 	print_c->setMaxLineSize(cfg_var_linelen.GetInt(cfg));
 }
 
+RAnnotatedCode* r2ghidra_decompile_annotated_code(RCore *core){
```
karliss, line 136: I would prefer if the internal implementation function passed offset explicitly instead of relying on global offset value. If necessary the core->offset can be passed as default position but that should be done at user interface level in the corresponding r2 command handler.

NirmalManoj: > I would prefer if the internal implementation function passed offset explicitly instead of relying on global offset value.

What benefit do we get from this? Is `static void Decompile(RCore *core, DecompileMode mode)` also not doing the same thing?

cutter-plugin/R2GhidraDecompiler.cpp
NirmalManoj: I have noticed this already. But the decompileAt() in R2Dec needs this, that's why I didn't change it here. If the core can be used in R2Dec::decompileAt(), we can avoid this here. This has to be decided.

src/core_ghidra.cpp
karliss: Yes there are other parts of code doing this. Poor separation of user interface from the implementation of main logic isn't uncommon in r2 and it makes it's API more difficult to use.

There are two aspects to this.

 One is that "global variables  are evil".  `core->offset`  isn't a one from language perspective but from the perspective of data-flow in the whole system it is one, with the same problems. You can find plenty of articles on this topic.

Other aspect is that core->offset is mixing responsibilities of 3 different concepts.
* core->offset as part of interface  - this only makes sense in the r2 command mode, in Cutter and Panel mode there is no single current offset.
* current offset as implementation detail of some io backends
* function capable of decompiling code at any point of program

Decompilation function shouldn't know anything about the current interface state. It is the responsibility of user interface tracking which parts you are looking at and to request decompiling them.

ITAYC0HEN: I am joining @karliss here with these concerns. 
@NirmalManoj please suggest better API layer functions to demonstrate the desired changes, show us and then implement if it looks good :)

NirmalManoj: Can we change `RAnnotatedCode* r2ghidra_decompile_annotated_code(RCore *core)` to `RAnnotatedCode* r2ghidra_decompile_annotated_code(RCore *core, RVA addr)`? 

With this change,  in `r2ghidra_decompile_annotated_code` we decompile code at the address `RVA addr` without using core->offset.  @karliss @ITAYC0HEN

NirmalManoj: I have made this change in the new commit @karliss @ITAYC0HEN . Now, cutter's plugin doesn't use core->offset directly or indirectly(as far as I could see:))

cutter-plugin/R2GhidraDecompiler.cpp
karliss: Looks good. Weird that I can't mark this comment as resolved.

src/core_ghidra.cpp
karliss: Looks good.

cutter-plugin/R2GhidraDecompiler.cpp
NirmalManoj: I will mark it as resolved:)

src/core_ghidra.cpp
NirmalManoj: @thestr4ng3r Should I make similar changes for retdec in the current PR?

thestr4ng3r: Yes please.

NirmalManoj: @thestr4ng3r Done. Please check if I have missed someplace that fetches offset from the core.

src/core_ghidra.cpp
```diff
@@ -132,6 +133,87 @@ static void ApplyPrintCConfig(RConfig *cfg, PrintC *print_c)
 	print_c->setMaxLineSize(cfg_var_linelen.GetInt(cfg));
 }
 
+RAnnotatedCode* r2ghidra_decompile_annotated_code(RCore *core, ut64 addr){
+	DecompilerLock lock;
+	RAnnotatedCode *code = nullptr;
+#ifndef DEBUG_EXCEPTIONS
```
karliss, line 139: I don't think it's necessary to keep old hacky debug code. Also the same can be easily achieved by setting breakpoint on exceptions or in the catch statement.

NirmalManoj: @karliss  Can you please look into the code and suggest parts you think I should separate? I had asked for suggestions to refactor for avoiding code duplication in the description of this PR.  I am not really sure about what all should be changed.

src/core_ghidra.cpp
```diff
@@ -271,7 +281,9 @@ static void Decompile(RCore *core, DecompileMode mode)
 		{
 			PJ *pj = pj_new ();
 			if(!pj)
+			{
```
thestr4ng3r, line 284: Please keep the code style as is, see https://gist.github.com/thestr4ng3r/d0b1de10852ca3d119e8c83375211e4b

src/core_ghidra.cpp
```diff
@@ -132,109 +133,120 @@ static void ApplyPrintCConfig(RConfig *cfg, PrintC *print_c)
 	print_c->setMaxLineSize(cfg_var_linelen.GetInt(cfg));
 }
 
-static void Decompile(RCore *core, DecompileMode mode)
-{
-	DecompilerLock lock;
-
-#ifndef DEBUG_EXCEPTIONS
+static void refactored_decompile(RCore *&core, std::stringstream &out_stream,
```
thestr4ng3r, line 136: Better signature:
```C++
static void Decompile(RCore *core, ut64 addr, DecompileMode mode, std::stringstream &out_stream, RAnnotatedCode **out_code)
```
The `&` for core is unnecessary, for code it's better readable like this (you will have to use `*out_code` instead of `code` in the function body).
Also this makes is more clear what the inputs and what the outputs are.

src/core_ghidra.cpp
```diff
@@ -132,109 +133,120 @@ static void ApplyPrintCConfig(RConfig *cfg, PrintC *print_c)
...
+		std::string s = "Ghidra Decompiler Error: " + error.explain;
+		char *err = strdup (s.c_str());
+ 		code = r_annotated_code_new(err);
+		// Push an annotation with: range = full string, type = error
+		// For this, we have to modify RAnnotatedCode to have one more type; for errors
+		return code;
+	}
+}
+
+static void Decompile(RCore *core, DecompileMode mode)
```
thestr4ng3r, line 242:
```suggestion
static void DecompileCmd(RCore *core, DecompileMode mode)
```
to distinguish it from the regular `Decompile()` as mentioned above.

src/core_ghidra.cpp
```diff
@@ -132,109 +133,120 @@ static void ApplyPrintCConfig(RConfig *cfg, PrintC *print_c)
 	print_c->setMaxLineSize(cfg_var_linelen.GetInt(cfg));
 }
 
-static void Decompile(RCore *core, DecompileMode mode)
-{
-	DecompilerLock lock;
-
-#ifndef DEBUG_EXCEPTIONS
+static void refactored_decompile(RCore *&core, std::stringstream &out_stream,
+								DecompileMode mode, RAnnotatedCode *&code, ut64 addr){
```
thestr4ng3r, line 137: Excessive indent, you can keep this all on a single line.

src/r2ghidra_annotated_code.h
```diff
@@ -0,0 +1,9 @@
+#ifndef R2GHIDRA_RANNOTATEDCODE_H
+#define R2GHIDRA_RANNOTATEDCODE_H
+
+#include <r_util/r_annotated_code.h>
+#include <r_core.h>
+
+R_API RAnnotatedCode* r2ghidra_decompile_annotated_code(RCore *core, ut64 addr);
```
thestr4ng3r, line 7:
```suggestion
R_API RAnnotatedCode *r2ghidra_decompile_annotated_code(RCore *core, ut64 addr);
```

NirmalManoj: Changed

src/core_ghidra.cpp
NirmalManoj: Changed as requested above.

src/core_ghidra.cpp
NirmalManoj: Done.

src/core_ghidra.cpp
NirmalManoj: Changed.

src/core_ghidra.cpp
```diff
@@ -260,9 +272,7 @@ static void Decompile(RCore *core, DecompileMode mode)
 				r_cons_printf("%s\n", out_stream.str().c_str());
 				break;
 		}
-
 		r_annotated_code_free(code);
-#ifndef DEBUG_EXCEPTIONS
```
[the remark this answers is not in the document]
NirmalManoj: Done.

CMakeLists.txt
```diff
@@ -46,7 +46,8 @@ set(CORE_SOURCE
 		src/R2PrintC.h
 		src/R2PrintC.cpp
 		src/RCoreMutex.h
-		src/RCoreMutex.cpp)
+		src/RCoreMutex.cpp
+		src/r2ghidra_annotated_code.h)
```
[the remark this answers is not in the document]
NirmalManoj: Done.

src/core_ghidra.cpp
NirmalManoj: @thestr4ng3r Please review again

ITAYC0HEN:
Woo-hoo! merged :)
Nirmal asked to "Merge all these three PRs together to ensure smooth functioning of all decompilers."

so we should quickly move on to the others, right? 
:)

PR #120, opened 2020-06-20
RAnal SLEIGH plugin

FXTi:
**Detailed description**

* Improve `pdgsd` result in RCore plugin.
* Port `SleighInstructionPrototype` from Ghidra's JAVA codebase to identify instruction's control flow type from underlying `Constructor`.
* Implement as-much-as-possible P-code to ESIL translation.
* Implement pattern-match on P-codes to identify type of single instruction.
* Improve processor and compiler spec file parse to extract reg info.
* Add `pdga` command to RCore plugin to enable this plugin.
* Enable plugin's compilation option as default.

**Test plan**

Has been tested on 8 cases in r2-testbins. Test cases will be added in next pr.

### DIFF
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -1,3 +1,5 @@
+/* radare - LGPL - Copyright 2020 - FXTi */
+
 #include "SleighAsm.h"
 
 void SleighAsm::init(RAsm *a)
@@ -56,32 +58,39 @@ void SleighAsm::initInner(RIO *io, char *cpu)
 	sleigh_id = cpu;
 }
 
-static void parseProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names) {
-	if (el->getName() != "prototype")
+static void parseProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names)
+{
+	if(el->getName() != "prototype")
 		throw LowlevelError("Expecting <prototype> tag");
 
 	const List &list(el->getChildren());
-	for(auto iter = list.begin(); iter != list.end(); ++iter) {
+	for(auto iter = list.begin(); iter != list.end(); ++iter)
+	{
 		const Element *subnode = *iter;
 
-		if (subnode->getName() == "input" || subnode->getName() == "output") {
-			// input->restoreXml(subnode,glb,effectlist,stackgrowsnegative);
+		if(subnode->getName() == "input" || subnode->getName() == "output")
+		{
 			const List &flist(el->getChildren());
-			for(auto fiter = flist.begin(); fiter != flist.end(); ++fiter) {
+			for(auto fiter = flist.begin(); fiter != flist.end(); ++fiter)
+			{
 				const Element *subel = *fiter;
 				const Element *reg = *subel->getChildren().begin();
-				if (subel->getName() == "pentry" && reg->getName() == "register") {
+				if(subel->getName() == "pentry" && reg->getName() == "register")
+				{
 					int4 num = subel->getNumAttributes(), i = 0;
-					for(; i < num; ++i) {
-						if (subel->getAttributeName(i) == "metatype" && subel->getAttributeValue(i) == "float")
+					for(; i < num; ++i)
+					{
+						if(subel->getAttributeName(i) == "metatype" && subel->getAttributeValue(i) == "float")
 							break;
 					}
-					if (i != num)
+					if(i != num)
 						continue;
 
-					for (int p = 0; p < reg->getNumAttributes(); ++p) {
-						if (reg->getAttributeName(p) == "name") {
-							if (subnode->getName() == "input")
+					for(int p = 0; p < reg->getNumAttributes(); ++p)
+					{
+						if(reg->getAttributeName(p) == "name")
+						{
+							if(subnode->getName() == "input")
 								arg_names.push_back(reg->getAttributeValue(p));
 							else
 								ret_names.push_back(reg->getAttributeValue(p));
@@ -93,11 +102,13 @@ static void parseProto(const Element *el, std::vector<std::string> &arg_names, s
 	}
 }
 
-static void parseDefaultProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names) {
+static void parseDefaultProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names)
+{
 	const List &list(el->getChildren());
 	List::const_iterator iter;
 
-	for(iter=list.begin();iter!=list.end();++iter) {
+	for(iter=list.begin();iter!=list.end();++iter)
+	{
 		// Decompiler will parse the same entry, and exit if multiple exists.
 		arg_names.clear();
 		ret_names.clear();
@@ -119,9 +130,8 @@ void SleighAsm::parseCompConfig(DocumentStorage &store)
 		const string &elname((*iter)->getName());
 		if(elname == "stackpointer")
 			sp_name = (*iter)->getAttributeValue("register");
-		if(elname == "default_proto") {
+		if(elname == "default_proto")
 			parseDefaultProto(*iter, arg_names, ret_names);
-		}
 	}
 }
 
@@ -188,9 +198,7 @@ void SleighAsm::parseProcConfig(DocumentStorage &store)
 			pc_name = (*iter)->getAttributeValue("register");
 
 		if(elname == "register_data")
-		{
 			reg_group = parseRegisterData(*iter);
-		}
 	}
 }
 
@@ -202,7 +210,7 @@ void SleighAsm::buildSpecfile(DocumentStorage &store)
 {
 	const LanguageDescription &language(description[languageindex]);
 	std::string compiler = sleigh_id.substr(sleigh_id.rfind(':')+1);
-	const CompilerTag &compilertag( language.getCompiler(compiler));
+	const CompilerTag &compilertag(language.getCompiler(compiler));
 
 	std::string processorfile;
 	std::string compilerfile;
@@ -461,7 +469,7 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 		auto *ins = trans.getInstruction(addr);
 		stringstream ss;
 		ss << assem.str << " " << ins->printFlowType(ins->getFlowType());
-		for (auto p : ins->getFlows())
+		for(auto p : ins->getFlows())
 			ss << " " << p;
 		r_strbuf_set(&op->buf_asm, ss.str().c_str());
 	}
@@ -514,8 +522,10 @@ std::vector<R2Reg> SleighAsm::getRegs(void)
 	return r2_reglist;
 }
 
-ostream &operator<<(ostream &s,const PcodeOperand &arg) {
-	switch (arg.type) {
+ostream &operator<<(ostream &s,const PcodeOperand &arg)
+{
+	switch(arg.type)
+	{
 		case PcodeOperand::REGISTER: s << arg.name; 
 			break;
 		case PcodeOperand::UNIQUE: s << "unique(" << arg.offset << ", " << arg.size << ")";
@@ -531,7 +541,8 @@ ostream &operator<<(ostream &s,const PcodeOperand &arg) {
 	return s;
 }
 
-ostream &operator<<(ostream &s,const Pcodeop &op) {
+ostream &operator<<(ostream &s,const Pcodeop &op)
+{
 	if (op.output)
 		s << *op.output << " = ";
 	s << get_opname(op.type);
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -416,17 +416,12 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 	return length;
 }
 
-int SleighAsm::genOpcode(RAnalOp *op, unsigned long long offset)
+int SleighAsm::genOpcode(PcodeSlg &pcode_slg, unsigned long long offset)
 {
-	if(!op)
-		throw LowlevelError("gen_opcode: op is nullptr.");
-
-	PcodeSlg pcode_slg(&trans);
 	Address addr(trans.getDefaultCodeSpace(), offset);
 	int length = 0;
 	try
 	{
-		/* RAnalOp *op is just to keep sync with disassemble() */
 		length = trans.oneInstruction(pcode_slg, addr);
 	}
 	catch(BadDataError &err)
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -58,7 +58,8 @@ void SleighAsm::initInner(RIO *io, char *cpu)
 	sleigh_id = cpu;
 }
 
-static void parseProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names)
+static void parseProto(const Element *el, std::vector<std::string> &arg_names,
+                       std::vector<std::string> &ret_names)
 {
 	if(el->getName() != "prototype")
 		throw LowlevelError("Expecting <prototype> tag");
@@ -80,7 +81,8 @@ static void parseProto(const Element *el, std::vector<std::string> &arg_names, s
 					int4 num = subel->getNumAttributes(), i = 0;
 					for(; i < num; ++i)
 					{
-						if(subel->getAttributeName(i) == "metatype" && subel->getAttributeValue(i) == "float")
+						if(subel->getAttributeName(i) == "metatype" &&
+						   subel->getAttributeValue(i) == "float")
 							break;
 					}
 					if(i != num)
@@ -102,12 +104,13 @@ static void parseProto(const Element *el, std::vector<std::string> &arg_names, s
 	}
 }
 
-static void parseDefaultProto(const Element *el, std::vector<std::string> &arg_names, std::vector<std::string> &ret_names)
+static void parseDefaultProto(const Element *el, std::vector<std::string> &arg_names,
+                              std::vector<std::string> &ret_names)
 {
 	const List &list(el->getChildren());
 	List::const_iterator iter;
 
-	for(iter=list.begin();iter!=list.end();++iter)
+	for(iter = list.begin(); iter != list.end(); ++iter)
 	{
 		// Decompiler will parse the same entry, and exit if multiple exists.
 		arg_names.clear();
@@ -209,7 +212,7 @@ void SleighAsm::parseProcConfig(DocumentStorage &store)
 void SleighAsm::buildSpecfile(DocumentStorage &store)
 {
 	const LanguageDescription &language(description[languageindex]);
-	std::string compiler = sleigh_id.substr(sleigh_id.rfind(':')+1);
+	std::string compiler = sleigh_id.substr(sleigh_id.rfind(':') + 1);
 	const CompilerTag &compilertag(language.getCompiler(compiler));
 
 	std::string processorfile;
@@ -217,7 +220,7 @@ void SleighAsm::buildSpecfile(DocumentStorage &store)
 	std::string slafile;
 
 	specpaths.findFile(processorfile, language.getProcessorSpec());
-	specpaths.findFile(compilerfile,compilertag.getSpec());
+	specpaths.findFile(compilerfile, compilertag.getSpec());
 	specpaths.findFile(slafile, language.getSlaFile());
 
 	try
@@ -240,15 +243,20 @@ void SleighAsm::buildSpecfile(DocumentStorage &store)
 		throw SleighError(serr.str());
 	}
 
-	try {
+	try
+	{
 		Document *doc = store.openDocument(compilerfile);
 		store.registerTag(doc->getRoot());
-	} catch(XmlError &err) {
+	}
+	catch(XmlError &err)
+	{
 		ostringstream serr;
 		serr << "XML error parsing compiler specification: " << compilerfile;
 		serr << "\n " << err.explain;
 		throw SleighError(serr.str());
-	} catch(LowlevelError &err) {
+	}
+	catch(LowlevelError &err)
+	{
 		ostringstream serr;
 		serr << "Error reading compiler specification: " << compilerfile;
 		serr << "\n " << err.explain;
@@ -316,7 +324,8 @@ void SleighAsm::scanSleigh(const string &rootpath)
 	FileManage::scanDirectoryRecursive(ghidradir, "Ghidra", rootpath, 2);
 	for(size_t i = 0; i < ghidradir.size(); ++i)
 	{
-		FileManage::scanDirectoryRecursive(procdir, "Processors", ghidradir[i], 1); // Look for Processors structure
+		FileManage::scanDirectoryRecursive(procdir, "Processors", ghidradir[i],
+		                                   1); // Look for Processors structure
 		FileManage::scanDirectoryRecursive(procdir, "contrib", ghidradir[i], 1);
 	}
 	if(procdir.size() != 0)
@@ -339,8 +348,8 @@ void SleighAsm::scanSleigh(const string &rootpath)
 		for(size_t i = 0; i < languagedirs.size(); ++i)
 			FileManage::directoryList(languagesubdirs, languagedirs[i]);
 	}
-	// If we haven't matched this directory structure, just use the rootpath as the directory containing
-	// the ldef
+	// If we haven't matched this directory structure, just use the rootpath as the directory
+	// containing the ldef
 	if(languagesubdirs.size() == 0)
 		languagesubdirs.push_back(rootpath);
 
@@ -469,7 +478,7 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 		auto *ins = trans.getInstruction(addr);
 		stringstream ss;
 		ss << assem.str << " " << ins->printFlowType(ins->getFlowType());
-		for(auto p : ins->getFlows())
+		for(auto p: ins->getFlows())
 			ss << " " << p;
 		r_strbuf_set(&op->buf_asm, ss.str().c_str());
 	}
@@ -516,39 +525,35 @@ std::vector<R2Reg> SleighAsm::getRegs(void)
 			sleigh_offset = p->first.offset;
 			sleigh_last = sleigh_offset + p->first.size;
 		}
-		r2_reglist.push_back(R2Reg{p->second, p->first.size, p->first.offset - sleigh_offset + offset});
+		r2_reglist.push_back(
+		    R2Reg{p->second, p->first.size, p->first.offset - sleigh_offset + offset});
 	}
 
 	return r2_reglist;
 }
 
-ostream &operator<<(ostream &s,const PcodeOperand &arg)
+ostream &operator<<(ostream &s, const PcodeOperand &arg)
 {
 	switch(arg.type)
 	{
-		case PcodeOperand::REGISTER: s << arg.name; 
-			break;
-		case PcodeOperand::UNIQUE: s << "unique(" << arg.offset << ", " << arg.size << ")";
-			break;
+		case PcodeOperand::REGISTER: s << arg.name; break;
+		case PcodeOperand::UNIQUE: s << "unique(" << arg.offset << ", " << arg.size << ")"; break;
 		// case PcodeOperand::RAM: s << "ram(" << arg.offset << ", " << arg.size << ")";
-		case PcodeOperand::RAM: s << arg.offset;
-			break;
-		case PcodeOperand::CONST: s << arg.number;
-			break;
-		default: 
-			throw LowlevelError("Unexpected type of PcodeOperand found in operator<<.");
+		case PcodeOperand::RAM: s << arg.offset; break;
+		case PcodeOperand::CONST: s << arg.number; break;
+		default: throw LowlevelError("Unexpected type of PcodeOperand found in operator<<.");
 	}
 	return s;
 }
 
-ostream &operator<<(ostream &s,const Pcodeop &op)
+ostream &operator<<(ostream &s, const Pcodeop &op)
 {
-	if (op.output)
+	if(op.output)
 		s << *op.output << " = ";
 	s << get_opname(op.type);
-	if (op.input0)
+	if(op.input0)
 		s << " " << *op.input0;
-	if (op.input1)
+	if(op.input1)
 		s << " " << *op.input1;
 	return s;
 }
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -34,6 +34,7 @@ void SleighAsm::initInner(RIO *io, const char *cpu)
 	parseCompConfig(docstorage);
 	alignment = trans.getAlignment();
 	trans.clearCache();
+	initRegMapping();
 
 	sleigh_id = cpu;
 }
@@ -448,7 +449,7 @@ std::string SleighAsm::getSleighHome(RConfig *cfg)
 
 int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 {
-	AssemblySlg assem;
+	AssemblySlg assem(this);
 	Address addr(trans.getDefaultCodeSpace(), offset);
 	int length = 0;
 	try
@@ -497,6 +498,25 @@ int SleighAsm::genOpcode(PcodeSlg &pcode_slg, Address &addr)
 		length = -1;
 	}
 	return length;
+}
+
+void SleighAsm::initRegMapping(void)
+{
+	reg_mapping.clear();
+	std::map<VarnodeData, std::string> reglist;
+	std::set<std::string> S;
+	trans.getAllRegisters(reglist);
+
+	for(auto iter = reglist.cbegin(); iter != reglist.cend(); ++iter)
+	{
+		std::string tmp;
+		for(auto p = iter->second.cbegin(); p != iter->second.cend(); ++p)
+			tmp.push_back(std::tolower(*p));
+		while(S.count(tmp))
+			tmp += "_dup";
+		S.insert(tmp);
+		reg_mapping[iter->second] = tmp;
+	}
 }
 
 std::vector<R2Reg> SleighAsm::getRegs(void)
@@ -549,4 +569,54 @@ ostream &operator<<(ostream &s, const Pcodeop &op)
 	if(op.input1)
 		s << " " << *op.input1;
 	return s;
+}
+
+void AssemblySlg::dump(const Address &addr, const string &mnem, const string &body)
+{
+	std::string res;
+	for (ut64 i = 0; i < body.size();)
+	{
+		std::string tmp;
+		while(!std::isalnum(body[i]))
+			res.push_back(body[i++]);
+		while(std::isalnum(body[i]))
+			tmp.push_back(body[i++]);
+		if(sasm->reg_mapping.find(tmp) != sasm->reg_mapping.end())
+			res += sasm->reg_mapping[tmp];
+		else
+			res += tmp;
+	}
+	str = r_str_newf("%s %s", mnem.c_str(), res.c_str());
+}
+
+PcodeOperand *PcodeSlg::parse_vardata(VarnodeData &data)
+{
+	AddrSpace *space = data.space;
+	PcodeOperand *operand = nullptr;
+	if(space->getName() == "register" || space->getName() == "mem")
+	{
+		operand = new PcodeOperand(
+		    sanal->reg_mapping[space->getTrans()->getRegisterName(data.space, data.offset, data.size)], data.size);
+		operand->type = PcodeOperand::REGISTER;
+	}
+	else if(space->getName() == "ram" || space->getName() == "DATA" || space->getName() == "code")
+	{
+		operand = new PcodeOperand(data.offset, data.size);
+		operand->type = PcodeOperand::RAM;
+	}
+	else if(space->getName() == "const")
+	{
+		// space.cc's ConstantSpace::printRaw()
+		operand = new PcodeOperand(data.offset);
+		operand->type = PcodeOperand::CONST;
+		operand->size = data.size; // To aviod ctor's signature collide with RAM's
+	}
+	else if(space->getName() == "unique")
+	{
+		operand = new PcodeOperand(data.offset, data.size);
+		operand->type = PcodeOperand::UNIQUE;
+	}
+	else
+		throw LowlevelError("Unsupported AddrSpace type appear.");
+	return operand;
 }
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -461,7 +461,7 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 		stringstream ss;
 		ss << assem.str << " " << ins->printFlowType(ins->getFlowType());
 		for(auto p: ins->getFlows())
-			ss << " " << p;
+		    ss << " " << p;
 		r_strbuf_set(&op->buf_asm, ss.str().c_str());
 		*/
 	}
@@ -574,7 +574,7 @@ ostream &operator<<(ostream &s, const Pcodeop &op)
 void AssemblySlg::dump(const Address &addr, const string &mnem, const string &body)
 {
 	std::string res;
-	for (ut64 i = 0; i < body.size();)
+	for(ut64 i = 0; i < body.size();)
 	{
 		std::string tmp;
 		while(!std::isalnum(body[i]))
@@ -595,8 +595,9 @@ PcodeOperand *PcodeSlg::parse_vardata(VarnodeData &data)
 	PcodeOperand *operand = nullptr;
 	if(space->getName() == "register" || space->getName() == "mem")
 	{
-		operand = new PcodeOperand(
-		    sanal->reg_mapping[space->getTrans()->getRegisterName(data.space, data.offset, data.size)], data.size);
+		operand = new PcodeOperand(sanal->reg_mapping[space->getTrans()->getRegisterName(
+		                               data.space, data.offset, data.size)],
+		                           data.size);
 		operand->type = PcodeOperand::REGISTER;
 	}
 	else if(space->getName() == "ram" || space->getName() == "DATA" || space->getName() == "code")
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -405,7 +405,13 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 	try
 	{
 		length = trans.printAssembly(assem, addr);
-		r_strbuf_set(&op->buf_asm, assem.str);
+		// r_strbuf_set(&op->buf_asm, assem.str);
+		auto *ins = trans.getInstruction(addr);
+		stringstream ss;
+		ss << assem.str << " " << ins->printFlowType(ins->getFlowType());
+		for (auto p : ins->getFlows())
+			ss << " " << p;
+		r_strbuf_set(&op->buf_asm, ss.str().c_str());
 	}
 	catch(BadDataError &err)
 	{
@@ -454,4 +460,31 @@ std::vector<R2Reg> SleighAsm::getRegs(void)
 	}
 
 	return r2_reglist;
+}
+
+ostream &operator<<(ostream &s,const PcodeOperand &arg) {
+	switch (arg.type) {
+		case PcodeOperand::REGISTER: s << arg.name; 
+			break;
+		case PcodeOperand::UNIQUE: s << "unique(" << arg.offset << ", " << arg.size << ")";
+			break;
+		case PcodeOperand::RAM: s << "ram(" << arg.offset << ", " << arg.size << ")";
+			break;
+		case PcodeOperand::CONST: s << arg.number;
+			break;
+		default: 
+			throw LowlevelError("Unexpected type of PcodeOperand found in operator<<.");
+	}
+	return s;
+}
+
+ostream &operator<<(ostream &s,const Pcodeop &op) {
+	if (op.output)
+		s << *op.output << " = ";
+	s << get_opname(op.type);
+	if (op.input0)
+		s << " " << *op.input0;
+	if (op.input1)
+		s << " " << *op.input1;
+	return s;
 }
--- a/src/SleighAsm.cpp
+++ b/src/SleighAsm.cpp
@@ -4,7 +4,7 @@ void SleighAsm::init(RAsm *a)
 {
 	if(description.empty())
 	{
-	/* Initialize sleigh spec files */
+		/* Initialize sleigh spec files */
 		scanSleigh(getSleighHome(getConfig(a)));
 		collectSpecfiles();
 	}
@@ -17,30 +17,98 @@ void SleighAsm::init(RAsm *a)
 	if(!io)
 		throw LowlevelError("Can't get RIO from RBin");
 
+	initInner(io, a->cpu);
+}
+
+void SleighAsm::init(RAnal *a)
+{
+	if(description.empty())
+	{
+		/* Initialize sleigh spec files */
+		scanSleigh(getSleighHome(getConfig(a)));
+		collectSpecfiles();
+	}
+
+	if(!sleigh_id.empty() && sleigh_id == std::string(a->cpu))
+		return;
+
+	RIO *io = a ? a->iob.io : nullptr;
+	if(!io)
+		throw LowlevelError("Can't get RIO from RBin");
+
+	initInner(io, a->cpu);
+}
+
+void SleighAsm::initInner(RIO *io, char *cpu)
+{
 	/* Initialize Sleigh */
 	loader = std::move(AsmLoadImage(io));
 	docstorage = std::move(DocumentStorage());
-	resolveArch(a->cpu);
+	resolveArch(cpu);
 	buildSpecfile(docstorage);
 	context = std::move(ContextInternal());
 	trans.reset(&loader, &context);
 	trans.initialize(docstorage);
 	parseProcConfig(docstorage);
-	parseAlignment(docstorage);
+	parseCompConfig(docstorage);
+	alignment = trans.getAlignment();
 
-	sleigh_id = a->cpu;
+	sleigh_id = cpu;
 }
 
-/* This function will be removed in upcoming RAnal PR */
-void SleighAsm::parseAlignment(DocumentStorage &doc)
+void SleighAsm::parseCompConfig(DocumentStorage &store)
 {
-	const Element *el = doc.getTag("sleigh");
-	if (!el)
-		throw LowlevelError("Could not find sleigh tag");
+	const Element *el = store.getTag("compiler_spec");
+	if(!el)
+		throw LowlevelError("No compiler configuration tag found");
 
-	istringstream s(el->getAttributeValue("align"));
-	s.unsetf(ios::dec | ios::hex | ios::oct);
-	s >> alignment;
+	const List &list(el->getChildren());
+	List::const_iterator iter;
+
+	for(iter = list.begin(); iter != list.end(); iter++)
+	{
+		const string &elname((*iter)->getName());
+		if(elname == "stackpointer")
+			sp_name = (*iter)->getAttributeValue("register");
+	}
+}
+
+static std::unordered_map<std::string, std::string> parseRegisterData(const Element *el)
+{
+	const List &child_list(el->getChildren());
+	List::const_iterator iter;
+
+	std::unordered_map<std::string, std::string> reg_group;
+
+	for(iter = child_list.begin(); iter != child_list.end(); iter++)
+	{
+		if((*iter)->getName() != "register")
+			throw LowlevelError("Unexpected node get from register_data in processor spec!");
+
+		const std::string &name = (*iter)->getAttributeValue("name");
+		std::string group, hidden, unused, rename;
+		try
+		{
+			group = (*iter)->getAttributeValue("group");
+			hidden = (*iter)->getAttributeValue("hidden");
+			unused = (*iter)->getAttributeValue("unused");
+			rename = (*iter)->getAttributeValue("rename");
+		}
+		catch(const XmlError &e)
+		{
+			std::string err_prefix("Unknown attribute: ");
+			if(e.explain == err_prefix + "group") { /* nothing */ }
+			else if(e.explain == err_prefix + "hidden") { /* nothing */ }
+			else if(e.explain == err_prefix + "unused") { /* nothing */ }
+			else if(e.explain == err_prefix + "rename") { /* nothing */ }
+			else
+				throw;
+		}
+
+		reg_group.insert({name, group});
+	}
+
+	return reg_group;
 }
 
 /*
@@ -63,26 +131,33 @@ void SleighAsm::parseProcConfig(DocumentStorage &store)
 		const string &elname((*iter)->getName());
 		if(elname == "context_data")
 			context.restoreFromSpec(*iter, &trans);
+
+		if(elname == "programcounter")
+			pc_name = (*iter)->getAttributeValue("register");
+
+		if(elname == "register_data")
+		{
+			reg_group = parseRegisterData(*iter);
+		}
 	}
 }
 
 /*
  * From sleigh_arch.cc's buildSpecFile()
  * This function is used to fill DocumentStorage with sleigh files.
- * It is stripped compiler specs, since Asm doesn't need them.
  */
 void SleighAsm::buildSpecfile(DocumentStorage &store)
 {
 	const LanguageDescription &language(description[languageindex]);
-	//std::string compiler = archid.substr(archid.rfind(':')+1);
-	//const CompilerTag &compilertag( language.getCompiler(compiler));
+	std::string compiler = sleigh_id.substr(sleigh_id.rfind(':')+1);
+	const CompilerTag &compilertag( language.getCompiler(compiler));
 
 	std::string processorfile;
-	//std::string compilerfile;
+	std::string compilerfile;
 	std::string slafile;
 
 	specpaths.findFile(processorfile, language.getProcessorSpec());
-	//specpaths.findFile(compilerfile,compilertag.getSpec());
+	specpaths.findFile(compilerfile,compilertag.getSpec());
 	specpaths.findFile(slafile, language.getSlaFile());
 
 	try
@@ -105,7 +180,6 @@ void SleighAsm::buildSpecfile(DocumentStorage &store)
 		throw SleighError(serr.str());
 	}
 
-	/*
 	try {
 		Document *doc = store.openDocument(compilerfile);
 		store.registerTag(doc->getRoot());
@@ -120,7 +194,6 @@ void SleighAsm::buildSpecfile(DocumentStorage &store)
 		serr << "\n " << err.explain;
 		throw SleighError(serr.str());
 	}
-	*/
 
 	try
 	{
@@ -273,8 +346,18 @@ RConfig *SleighAsm::getConfig(RAsm *a)
 	return core->config;
 }
 
+RConfig *SleighAsm::getConfig(RAnal *a)
+{
+	RCore *core = a ? (RCore *)a->coreb.core : nullptr;
+	if(!core)
+		throw LowlevelError("Can't get RCore from RAnal's RCoreBind");
+	return core->config;
+}
+
 std::string SleighAsm::getSleighHome(RConfig *cfg)
 {
+	if(!cfg)
+		throw LowlevelError("SleighAsm::get_sleigh_home: cfg is nullptr.");
 	const char varname[] = "r2ghidra.sleighhome";
 	const char *path = nullptr;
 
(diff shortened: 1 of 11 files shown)


FXTi:
> How to label a bunch of P-codes generated from one single processor instruction.

For example, `SAR RSI,0x3` should be labeled as `R_ANAL_OP_TYPE_SAR`.

But in P-codes, it's :
```asm
0x00000725: SAR RSI,0x3
    (unique,0xa420,4) = INT_AND 0x3, 0x3f
    (unique,0xa430,8) = COPY RSI
    RSI = INT_SRIGHT RSI, (unique,0xa420,4)
    (unique,0x2aa0,1) = INT_NOTEQUAL (unique,0xa420,4), 0x0
    (unique,0x2ab0,4) = INT_SUB (unique,0xa420,4), 0x1
    (unique,0x2ac0,8) = INT_SRIGHT (unique,0xa430,8), (unique,0x2ab0,4)
    (unique,0x2ad0,8) = INT_AND (unique,0x2ac0,8), 0x1
    (unique,0x2af0,1) = INT_NOTEQUAL (unique,0x2ad0,8), 0x0
    (unique,0x2b00,1) = BOOL_NEGATE (unique,0x2aa0,1)
    (unique,0x2b10,1) = INT_AND (unique,0x2b00,1), CF
    (unique,0x2b20,1) = INT_AND (unique,0x2aa0,1), (unique,0x2af0,1)
    CF = INT_OR (unique,0x2b10,1), (unique,0x2b20,1)
    (unique,0x2b50,1) = INT_EQUAL (unique,0xa420,4), 0x1
    (unique,0x2b60,1) = BOOL_NEGATE (unique,0x2b50,1)
    OF = INT_AND (unique,0x2b60,1), OF
    (unique,0x2590,1) = INT_NOTEQUAL (unique,0xa420,4), 0x0
    (unique,0x25b0,1) = INT_SLESS RSI, 0x0
    (unique,0x25c0,1) = BOOL_NEGATE (unique,0x2590,1)
    (unique,0x25d0,1) = INT_AND (unique,0x25c0,1), SF
    (unique,0x25e0,1) = INT_AND (unique,0x2590,1), (unique,0x25b0,1)
    SF = INT_OR (unique,0x25d0,1), (unique,0x25e0,1)
    (unique,0x2610,1) = INT_EQUAL RSI, 0x0
    (unique,0x2620,1) = BOOL_NEGATE (unique,0x2590,1)
    (unique,0x2630,1) = INT_AND (unique,0x2620,1), ZF
    (unique,0x2640,1) = INT_AND (unique,0x2590,1), (unique,0x2610,1)
    ZF = INT_OR (unique,0x2630,1), (unique,0x2640,1)
    (unique,0x2660,1) = POPCOUNT RSI
    PF = INT_AND (unique,0x2660,1), 0x1
```

My solution here is:
1. Get operands from disas result, which is `RSI` and 0x3
2. filter register name from result above, which is `RSI`
3. Use register name to filter out the P-codes containing register name in operands, which are: 
```asm
  2     (unique,0xa430,8) = COPY RSI
  3     RSI = INT_SRIGHT RSI, (unique,0xa420,4)
  4     (unique,0x25b0,1) = INT_SLESS RSI, 0x0
  5     (unique,0x2610,1) = INT_EQUAL RSI, 0x0
  6     (unique,0x2660,1) = POPCOUNT RSI
```
4. Use only operator from result above, which are: `COPY INT_SRIGHT INT_SLESS INT_EQUAL POPCOUNT`
5. label those P-code operators and take the one matters most. which is `INT_SRIGHT -> R_ANAL_OP_TYPE_SAR`

FXTi:
> - What is the exact info that you get in Java-Ghidra about a single instruction? Maybe they already have more than just pcode somewhere, either coming from sleigh directly or derived from pcode.

About pcode info getting from SLEIGH, there're `Architecture::inst`, which is built in `Architecture::buildInstructions`. By calling `TypeOp::registerInstructions`, it actually set `opflags` and `behave`. And all these info is used by Emulate system. There's also `PcodeEmitCache` to export pcode from SLEIGH for Emulate system. But I think this could be equal usage but much heavier than my impl. I just need to do the pair, filter, and sort work.

> - What analysis/filtering operations are already provided by the decompiler/sleigh library? Perhaps you can use some kind of data flow analysis/filtering to mask the ops you want.

It's mainly data flow analysis, I think it doesn't fit.

thestr4ng3r:
> > * What is the exact info that you get in Java-Ghidra about a single instruction? Maybe they already have more than just pcode somewhere, either coming from sleigh directly or derived from pcode.
> 
> About pcode info getting from SLEIGH, there're `Architecture::inst`, which is built in `Architecture::buildInstructions`. By calling `TypeOp::registerInstructions`, it actually set `opflags` and `behave`. And all these info is used by Emulate system. There's also `PcodeEmitCache` to export pcode from SLEIGH for Emulate system. But I think this could be equal usage but much heavier than my impl. I just need to do the pair, filter, and sort work.

As far as I can tell, `TypeOp` are the pcode ops, not the real instructions. What I meant is what exactly do you get for e.g. a single x86 instruction aside from pcode. Also, did you check in Java yet?

XVilka:
@trufae @ret2libc by the way, please take a look when you have time and if you have any suggestions/complaints - write it here.

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		buf << p->name << "\t." << p->size * 8 << "\t" << p->offset << "\t" << "0\n";
+	}
+	const std::string &res = buf.str();
+	//fprintf(stderr, res.c_str());
+	return strdup(res.c_str());
+}
+
+RAnalPlugin r_anal_plugin_ghidra = {
```
XVilka, line 136: Why just not use designated initializers instead of comments?

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,179 @@
...
+	/* .esil_trap = */ nullptr,
+	/* .esil_fini = */ nullptr,
+};
+
+#ifndef CORELIB
+#ifdef __cplusplus
+extern "C"
+#endif
+R_API RLibStruct radare_plugin = {
+	/* .type = */ R_LIB_TYPE_ANAL,
```
XVilka, line 171: Same here.

src/SleighInstruction.h
```diff
@@ -0,0 +1,428 @@
...
+class OpTplWalker {
+	private:
+		ConstructState *point = nullptr;		// The current node being visited
+		const vector<OpTpl *> *oparray = nullptr;		// current array of ops being traversed
+		int4 depth;					// Depth of current node within the tree
+		int4 breadcrumb[64];			// Path of operands from the root
+		int maxsize;				// Maximum number of directives for this point
+		int sectionnum;
+
+		void setupPoint() {
```
XVilka, line 177: What about implementing the methods not in the header?

src/anal_ghidra.cpp
thestr4ng3r: They are not available in C++11.

XVilka: It's a middle of 2020 though...

thestr4ng3r: Tell that to certain distributions 🙈

FXTi: lol

FXTi:
The current issue is: ***hard to describe behaviour of R_ANAL_OP_TYPE_XXX across multi-arch***

Still take `SAR` instruction as example, I've implemented p-code filter function, and also consider middle variables, since middle variables can hold value from desired registers. 

About middle variables:
* It's casual to split one side affect of instruction into multiple P-code, and that introduce some variables holds middle value.
* SLEIGH use unique varnode for that.
* Those unique varnode will hold middle value and participate in later computation.

For example:
```
0x0000105c: PUSH EAX
    (unique,0x12f0,4) = COPY EAX
    ESP = INT_SUB ESP, 0x4
    (unique,0x12f0,4) = STORE [ESP]
```
It's obvious that middle variables(unique varnodes) is necessary for analysis.

Let's continue on original topic:

Raw P-code:
```
0x000011a3: SAR ESI,0xfd
    (unique,0x7900,4) = INT_AND 0xfd, 0x1f
    (unique,0x7910,4) = COPY ESI
    ESI = INT_SRIGHT ESI, (unique,0x7900,4)
    (unique,0x2060,1) = INT_NOTEQUAL (unique,0x7900,4), 0x0
    (unique,0x2070,4) = INT_SUB (unique,0x7900,4), 0x1
    (unique,0x2080,4) = INT_SRIGHT (unique,0x7910,4), (unique,0x2070,4)
    (unique,0x2090,4) = INT_AND (unique,0x2080,4), 0x1
    (unique,0x20b0,1) = INT_NOTEQUAL (unique,0x2090,4), 0x0
    (unique,0x20c0,1) = BOOL_NEGATE (unique,0x2060,1)
    (unique,0x20d0,1) = INT_AND (unique,0x20c0,1), CF
    (unique,0x20e0,1) = INT_AND (unique,0x2060,1), (unique,0x20b0,1)
    CF = INT_OR (unique,0x20d0,1), (unique,0x20e0,1)
    (unique,0x2110,1) = INT_EQUAL (unique,0x7900,4), 0x1
    (unique,0x2120,1) = BOOL_NEGATE (unique,0x2110,1)
    OF = INT_AND (unique,0x2120,1), OF
    (unique,0x1b50,1) = INT_NOTEQUAL (unique,0x7900,4), 0x0
    (unique,0x1b70,1) = INT_SLESS ESI, 0x0
    (unique,0x1b80,1) = BOOL_NEGATE (unique,0x1b50,1)
    (unique,0x1b90,1) = INT_AND (unique,0x1b80,1), SF
    (unique,0x1ba0,1) = INT_AND (unique,0x1b50,1), (unique,0x1b70,1)
    SF = INT_OR (unique,0x1b90,1), (unique,0x1ba0,1)
    (unique,0x1bd0,1) = INT_EQUAL ESI, 0x0
    (unique,0x1be0,1) = BOOL_NEGATE (unique,0x1b50,1)
    (unique,0x1bf0,1) = INT_AND (unique,0x1be0,1), ZF
    (unique,0x1c00,1) = INT_AND (unique,0x1b50,1), (unique,0x1bd0,1)
    ZF = INT_OR (unique,0x1bf0,1), (unique,0x1c00,1)
    (unique,0x1c20,1) = POPCOUNT ESI
    PF = INT_AND (unique,0x1c20,1), 0x1
```

Filtered P-code:
```
0x11a3: ESI = INT_SRIGHT ESI unique(7900, 4)
0x11a3: unique(2080, 4) = INT_SRIGHT unique(7910, 4) unique(2070, 4)
0x11a3: unique(2090, 4) = INT_AND unique(2080, 4) 1
0x11a3: unique(20b0, 1) = INT_NOTEQUAL unique(2090, 4) 0
0x11a3: unique(20e0, 1) = INT_AND unique(2060, 1) unique(20b0, 1)
0x11a3: CF = INT_OR unique(20d0, 1) unique(20e0, 1)
0x11a3: unique(1b70, 1) = INT_SLESS ESI 0
0x11a3: unique(1ba0, 1) = INT_AND unique(1b50, 1) unique(1b70, 1)
0x11a3: SF = INT_OR unique(1b90, 1) unique(1ba0, 1)
0x11a3: unique(1bd0, 1) = INT_EQUAL ESI 0
0x11a3: unique(1c00, 1) = INT_AND unique(1b50, 1) unique(1bd0, 1)
0x11a3: ZF = INT_OR unique(1bf0, 1) unique(1c00, 1)
0x11a3: unique(1c20, 1) = POPCOUNT ESI
0x11a3: PF = INT_AND unique(1c20, 1) 1
0x11a3: unique(7910, 4) = COPY ESI
```

The filtered P-code is enough to know this instruction is doing `INT_SRIGHT`. But if we want to fill RAnal data structure like other RAnal plugins do:

```c
	case R_ANAL_OP_TYPE_SAR:
		CREATE_SRC_DST (op);
		set_src_dst (a->reg, op->dst, handle, insn, 0);
		set_src_dst (a->reg, op->src[0], handle, insn, 1);
		set_src_dst (a->reg, op->src[1], handle, insn, 2);
		set_src_dst (a->reg, op->src[2], handle, insn, 3);
```

It will require forward and backward lookup on complete P-codes to trace what unique varnodes originally are: a register, an address or a const value? And by this, I think of a new way to do identity work: model behaviour of R_ANAL_OP_TYPE_XXX across multi-arch and match P-codes with this pattern.

Still take `SAR` as example:
> SAR is applied on one victim register, doing signed right shift on victim register, and store back into victim register

And that can be translate into three constraints:
* P-codes must contain INT_SRIGHT
* the input0 varnode should be register or unique varnode generated from operation on register(input0 is varnode input being shifted)
* the output varnode should be shifted register or unique varnode which will be stored in shifted register

If all three constraints matches, we can say this instruction is `R_ANAL_OP_TYPE_SAR`.
But I don't know how related RAnal data structure should be filled and how accurate this solution will provide. Most importantly, whether my constraints are correct across multi-arches?

Moreover, I think ESIL is easier to translate from P-code. So, can I just provide accurate P-code and ignore related data structure?(Just label instructions) WIll radare2 extract necessary info from ESIL?

XVilka:
A lot of radare2 analysis steps requires this RAnalOp information rather than pure ESIL. Thus the plain P-code to ESIL translation wouldn't be enough. I suggest to make the plagin to fill some kind of hashtable to precompute these references, on the stage of initializing the analysis.

wargio:
I agree with @XVilka i do not expect to have SLEIGH handling perfectly the analysis data that is given to RAnal.

I would also suggest to precompute and stay away from ESIL linked analysis.

Maybe you can try with a simplified view of the issue (you know, to start) and then try to patch things out; is not the best solution, but since the problem is quite complex, i would start this way.

FXTi:
```c
	R_ANAL_OP_TYPE_MOV   = 9, /* register move */
	R_ANAL_OP_TYPE_CMOV  = 9 | R_ANAL_OP_TYPE_COND, /* conditional move */
	R_ANAL_OP_TYPE_TRAP  = 10, /* it's a trap! */
	R_ANAL_OP_TYPE_SWI   = 11,  /* syscall, software interrupt */
	R_ANAL_OP_TYPE_CSWI  = 11 | R_ANAL_OP_TYPE_COND,  /* syscall, software interrupt */
	R_ANAL_OP_TYPE_UPUSH = 12, /* unknown push of data into stack */
	R_ANAL_OP_TYPE_RPUSH = R_ANAL_OP_TYPE_UPUSH | R_ANAL_OP_TYPE_REG, /* push register */
	R_ANAL_OP_TYPE_PUSH  = 13,  /* push value into stack */
	R_ANAL_OP_TYPE_POP   = 14,   /* pop value from stack to register */
	R_ANAL_OP_TYPE_CMP   = 15,  /* compare something */
	R_ANAL_OP_TYPE_ACMP  = 16,  /* compare via and */
	R_ANAL_OP_TYPE_ADD   = 17,
	R_ANAL_OP_TYPE_SUB   = 18,
	R_ANAL_OP_TYPE_IO    = 19,
	R_ANAL_OP_TYPE_MUL   = 20,
	R_ANAL_OP_TYPE_DIV   = 21,
	R_ANAL_OP_TYPE_SHR   = 22,
	R_ANAL_OP_TYPE_SHL   = 23,
	R_ANAL_OP_TYPE_SAL   = 24,
	R_ANAL_OP_TYPE_SAR   = 25,
	R_ANAL_OP_TYPE_OR    = 26,
	R_ANAL_OP_TYPE_AND   = 27,
	R_ANAL_OP_TYPE_XOR   = 28,
	R_ANAL_OP_TYPE_NOR   = 29,
	R_ANAL_OP_TYPE_NOT   = 30,
	R_ANAL_OP_TYPE_STORE = 31,  /* store from register to memory */
	R_ANAL_OP_TYPE_LOAD  = 32,  /* load from memory to register */
	R_ANAL_OP_TYPE_LEA   = 33, /* TODO add ulea */
	R_ANAL_OP_TYPE_LEAVE = 34,
	R_ANAL_OP_TYPE_ROR   = 35,
	R_ANAL_OP_TYPE_ROL   = 36,
	R_ANAL_OP_TYPE_XCHG  = 37,
	R_ANAL_OP_TYPE_MOD   = 38,
	R_ANAL_OP_TYPE_SWITCH = 39,
	R_ANAL_OP_TYPE_CASE = 40,
	R_ANAL_OP_TYPE_LENGTH = 41,
	R_ANAL_OP_TYPE_CAST = 42,
	R_ANAL_OP_TYPE_NEW = 43,
	R_ANAL_OP_TYPE_ABS = 44,
	R_ANAL_OP_TYPE_CPL = 45,	/* complement */
	R_ANAL_OP_TYPE_CRYPTO = 46,
	R_ANAL_OP_TYPE_SYNC = 47,
```

Pattern needed. Good night.

wargio:
> It's recognize as `COMPUTED_CALL` due to `CALLIND`. Do I need to specify this is `R_ANAL_OP_TYPE_SWI`?
> 

No, i do not think is needed.

> Is `INT_ADD` and `INT_SUB` to include all situations?
> 

A CMP is always a sub.

> Isn't this equal to `R_ANAL_OP_TYPE_SAL`?

Hmm yes.

> ## R_ANAL_OP_TYPE_LEAVE = 34,
> 
> Attention:
> Recommend just ignore this. Is this acceptable?
> 
> Example:
> 
> ```assembly
> 0x000011ca: LEAVE 
>     ESP = COPY EBP
>     EBP = LOAD [ESP]
>     ESP = INT_ADD ESP, 0x4
> ```
> 

Yes, ignore it.

> ## R_ANAL_OP_TYPE_MOD = 38,
> 
> Pattern:
> 
>     * Has `INT_SREM` or `INT_REM`
> 
> 
> Attention:
> 
>     * Will this pattern accurate enough?
> 
>     * Unluckily, assembly need improvement. P-code result is right, `v1` and `v0` is dalvik's local variables.
> 
> 
> Example:
> 
> ```assembly
> 0x0000021e: rem_int_2addr 0x1004,0x1000
>     v1 = INT_SREM v1, v0
> ```

Yes, `R_ANAL_OP_TYPE_MOD` is signed and unsigned, therefore is correct your assumption.

> Attention:
> Sample binary file needed!

Have you checked in r2r-bins repo?

thestr4ng3r:
Fantastic overview!

> > It's recognize as `COMPUTED_CALL` due to `CALLIND`. Do I need to specify this is `R_ANAL_OP_TYPE_SWI`?
> 
> No, i do not think is needed.

I think it is needes because syscalls are very different instructions from calls. Since you already identified that they always have this `CALLOTHER`, do you think it would be hard to differenciate them?

> ## R_ANAL_OP_TYPE_STORE = 31, /* store from register to memory */
> 
> Attention:
> I don't know how to identify this from `mov` or others.

Store always has the destination in memory. I think if you encounter that it makes more sense to mark it as store than as mov. So having mov to memory marked as store is better than having store marked as mov in my opinion.

Essentially movs to memory are semantically just stores with a different name.

> ## R_ANAL_OP_TYPE_LOAD = 32, /* load from memory to register */
> 
> Attention:
> I don't know how to identify this from `mov` or others.

Same here, better mark movs as loads than the other way around.

> ## R_ANAL_OP_TYPE_LEA = 33, /* TODO add ulea */
> 
> Attention:
> I don't know how to identify this.

`lea` is x86 specific and essentially just a computation with a different name. I don't think you have to recognize them.



> ## R_ANAL_OP_TYPE_NOR = 29,
> 
> Attention:
> Sample binary file needed!

`radare2/test/bins/mach0/ppc-ls`:
```
            0x00005758      7f7bd8f8       nor r27, r27, r27
```
(or `r2 -appc -b32 -ecfg.bigendian=1 -c "wx 7f7bd8f8" -`)


> ## R_ANAL_OP_TYPE_CPL = 45, /* complement */
> 
> Attention:
> Sample binary file needed!
> ## R_ANAL_OP_TYPE_CRYPTO = 46,

`r2 -apic -easm.cpu=pic18 -c "pd1@0x300002" ihex:///home/florian/dev/radare2/test/bins/pic18c/FreeRTOS-pic18c.hex`
```
            0x00300002      181e           comf 0x18, 1, 0
```
or `r2 -apic -easm.cpu=pic18 -c "wx 181e" -`

However most plugins mark CPL instructions as NOT. Complement is just bitwise negation so having both seems redundant unless NOT is only meant for something like `dst = !src` where dst will be only 1 or 0 but that's not documented anywhere so I would just ignore CPL and mark all as NOT.

> Attention:
> Sample binary file needed!
> ## R_ANAL_OP_TYPE_SYNC = 47,
> 
> Attention:
> Sample binary file needed!

I didn't find a binary with this one, but got this by bruteforcing with random bytes:

`r2 -aarm -b64 -c "wx df3b03d5" -`
```
            0x00000000      df3b03d5       isb 0xb
```

FXTi:
P-code list for your reference: https://ghidra.re/courses/languages/html/pcodedescription.html

## R_ANAL_OP_TYPE_MOV   = 9, /* register move */

Pattern:
* Has `COPY`
* Input and output come from registers appear in assembly

Example:
```asm
0x00001191: MOV EBP,ESP
    EBP = COPY ESP
0x00001194: MOV dword ptr [EBP + -0x4],0x0
    (unique,0x3a0,4) = INT_ADD EBP, 0xfffffffc
    (unique,0xf40,4) = COPY 0x0
    (unique,0xf40,4) = STORE [(unique,0x3a0,4)]
0x0000119b: MOV dword ptr [EAX + -0x64],EAX
    (unique,0x3a0,4) = INT_ADD EAX, 0xffffff9c
    (unique,0xf30,4) = COPY EAX
    (unique,0xf30,4) = STORE [(unique,0x3a0,4)]
```

## R_ANAL_OP_TYPE_CMOV  = 9 | R_ANAL_OP_TYPE_COND, /* conditional move */

Pattern:
* Has `COPY`
* Input and output come from registers appear in assembly
* Has `CBRANCH` above it jumps to next instruction or jumps to this `COPY` or `STORE` pcode

Example:
```asm
0x000011a1: CMOVL ECX,dword ptr [0x3]
    (unique,0x1090,1) = INT_NOTEQUAL OF, SF
    (unique,0x3aa0,1) = BOOL_NEGATE (unique,0x1090,1)
    CBRANCH dword_ptr(0x000011a8), (unique,0x3aa0,1)
    ECX = COPY dword_ptr(0x00000003)
```

## R_ANAL_OP_TYPE_STORE = 31,  /* store from register to memory */

Pattern:
* Has `STORE`
* Input come from registers appear in assembly
* Output doesn't come from registers appear in assembly

Example:
```asm
0x00001218: STOSD ES:EDI
    (unique,0xe90,4) = COPY EDI
    (unique,0xea0,4) = INT_ADD EDI, 0x4
    (unique,0xeb0,4) = INT_ZEXT DF
    (unique,0xec0,4) = INT_MULT 0x8, (unique,0xeb0,4)
    EDI = INT_SUB (unique,0xea0,4), (unique,0xec0,4)
    (unique,0xee0,4) = COPY EAX
    (unique,0xee0,4) = STORE [(unique,0xe90,4)]
```

## R_ANAL_OP_TYPE_LOAD  = 32,  /* load from memory to register */

Pattern:
* Has `STORE`
* Input doesn't come from registers appear in assembly
* Output come from registers appear in assembly

Example:
```asm
0x00001218: LODSD ESI
    (unique,0xc20,4) = COPY ESI
    (unique,0xc30,4) = INT_ADD ESI, 0x4
    (unique,0xc40,4) = INT_ZEXT DF
    (unique,0xc50,4) = INT_MULT 0x8, (unique,0xc40,4)
    ESI = INT_SUB (unique,0xc30,4), (unique,0xc50,4)
    (unique,0xc70,4) = LOAD [(unique,0xc20,4)]
    EAX = COPY (unique,0xc70,4)
```

## R_ANAL_OP_TYPE_TRAP  = 10, /* it's a trap! */

No, there's no trap in raw pcode

## R_ANAL_OP_TYPE_SWI   = 11,  /* syscall, software interrupt */

Pattern:
* Has `CALLOTHER`

Attention:
This kind has higher priority than flow type analysis. Should specify this as `R_ANAL_OP_TYPE_SWI` first.
Because it's recognize as `COMPUTED_CALL` due to `CALLIND`

Example:
```asm
0x000011a8: INT 0x80
    (unique,0x5370,4) = CALLOTHER 0x10, 0x80
    CALLIND (unique,0x5370,4)

0x00400114: syscall 
    CALLOTHER 0x3, 0x0
```

## R_ANAL_OP_TYPE_CSWI  = 11 | R_ANAL_OP_TYPE_COND,  /* syscall, software interrupt */

Pattern:
* Has `CALLOTHER` and `CBRANCH`

Attention:
As above.

## R_ANAL_OP_TYPE_UPUSH = 12, /* unknown push of data into stack */

Pattern:
* Has `STROE`
* Output of `STORE` is or comes from `stack_reg`

Example:
```asm
0x00001190: PUSH EBP
    (unique,0x12f0,4) = COPY EBP
    ESP = INT_SUB ESP, 0x4
    (unique,0x12f0,4) = STORE [ESP]
```

## R_ANAL_OP_TYPE_RPUSH = R_ANAL_OP_TYPE_UPUSH | R_ANAL_OP_TYPE_REG, /* push register */

Pattern:
* Has `STROE`
* Output of `STORE` is or comes from `stack_reg`
* Input of `STORE` is or comes from a register appearing in assembly

Example:
```asm
0x00001190: PUSH EBP
    (unique,0x12f0,4) = COPY EBP
    ESP = INT_SUB ESP, 0x4
    (unique,0x12f0,4) = STORE [ESP]
```

## R_ANAL_OP_TYPE_PUSH  = 13,  /* push value into stack */

Pattern:
* Has `STROE`
* Output of `STORE` is or comes from `stack_reg`
* Input of `STORE` isn't a register
* Input of `STORE` doesn't come from a register

Example:
```asm
0x00001190: PUSH EBP
    (unique,0x12f0,4) = COPY EBP
    ESP = INT_SUB ESP, 0x4
    (unique,0x12f0,4) = STORE [ESP]
```

## R_ANAL_OP_TYPE_POP   = 14,   /* pop value from stack to register */

Pattern:
* Has `LOAD`
* Input of `LOAD` is or comes from `stack_reg`
* Output of `LOAD` is or comes from register appearing in assembly

Example:
```asm
0x000011af: POP EBP
    EBP = LOAD [ESP]
    ESP = INT_ADD ESP, 0x4
```

## R_ANAL_OP_TYPE_CMP   = 15,  /* compare something */

Pattern:
* Has `INT_SUB`
* Two inputs of `INT_SUB` is a register or comes from register appearing in assembly
* Output of key pcode is one input of `INT_EQUAL` or `INT_LESS` or `INT_NOTEQUAL` or `INT_LESSEQUAL` or `INT_SLESS` or `INT_SLESSEQUAL`

Example:
```asm
0x0000120b: CMP EBX,EAX
    CF = INT_LESS EBX, EAX
    OF = INT_SBORROW EBX, EAX
    (unique,0x3c00,4) = INT_SUB EBX, EAX
    SF = INT_SLESS (unique,0x3c00,4), 0x0
    ZF = INT_EQUAL (unique,0x3c00,4), 0x0
    (unique,0x1b20,1) = POPCOUNT (unique,0x3c00,4)
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_ACMP  = 16,  /* compare via and */

Pattern:
* Has `INT_AND`
* Two inputs of `INT_AND` is a register or comes from register appearing in assembly
* Output of key pcode is one input of `INT_EQUAL` or `INT_LESS` or `INT_NOTEQUAL` or `INT_LESSEQUAL` or `INT_SLESS` or `INT_SLESSEQUAL`

Example:
```asm
0x0000120b: TEST EBX,EAX
    CF = COPY 0x0
    OF = COPY 0x0
    (unique,0x83c0,4) = INT_AND EBX, EAX
    SF = INT_SLESS (unique,0x83c0,4), 0x0
    ZF = INT_EQUAL (unique,0x83c0,4), 0x0
    (unique,0x1b20,1) = POPCOUNT (unique,0x83c0,4)
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_ADD   = 17,

Pattern:
* Has `INT_ADD`
* Input of `INT_ADD` is a register or comes from register appearing in assembly
* Output of `INT_ADD` is a register or comes from register appearing in assembly

Example:
```asm
0x000011f4: ADD EBX,0x2e0c
    CF = INT_CARRY EBX, 0x2e0c
    OF = INT_SCARRY EBX, 0x2e0c
    EBX = INT_ADD EBX, 0x2e0c
    SF = INT_SLESS EBX, 0x0
    ZF = INT_EQUAL EBX, 0x0
    (unique,0x1b20,1) = POPCOUNT EBX
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_SUB   = 18,

Pattern:
* Has `INT_ADD`
* Input of `INT_ADD` is a register or comes from register appearing in assembly
* Output of `INT_ADD` is a register or comes from register appearing in assembly

Example:
```asm
0x000011ec: SUB ESP,0x10
    CF = INT_LESS ESP, 0x10
    OF = INT_SBORROW ESP, 0x10
    ESP = INT_SUB ESP, 0x10
    SF = INT_SLESS ESP, 0x0
    ZF = INT_EQUAL ESP, 0x0
    (unique,0x1b20,1) = POPCOUNT ESP
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_IO    = 19,

Attention:
I think this could be regarded as a syscall. So ignore this.

Example:
```asm
0x00001198: IN AL, 0x7f
    AL = CALLOTHER 0x1, 0x7f
```

## R_ANAL_OP_TYPE_MUL   = 20,

Pattern:
* Has `INT_MULT`
* Input of `INT_MULT` is a register or comes from register appearing in assembly
* Output of `INT_MULT` is a register or comes from register appearing in assembly

Example:
```asm
0x0000120b: IMUL EDX
    (unique,0x4e80,8) = INT_SEXT EAX
    (unique,0x4e90,8) = INT_SEXT EDX
    (unique,0x4eb0,8) = INT_MULT (unique,0x4e80,8), (unique,0x4e90,8)
    EDX = SUBPIECE (unique,0x4eb0,8), 0x4
    EAX = SUBPIECE (unique,0x4eb0,8), 0x0
    (unique,0x1c80,1) = INT_NOTEQUAL EDX, 0x0
    (unique,0x1c90,4) = INT_2COMP 0x1
    (unique,0x1ca0,1) = INT_NOTEQUAL EDX, (unique,0x1c90,4)
    CF = INT_AND (unique,0x1c80,1), (unique,0x1ca0,1)
    OF = COPY CF
```

## R_ANAL_OP_TYPE_DIV   = 21,

Pattern:
* Has `INT_DIV`
* Input of `INT_DIV` is a register or comes from register appearing in assembly
* Output of `INT_DIV` is a register or comes from register appearing in assembly

Example:
```asm
0x00001209: DIV EDX
    (unique,0x45c0,8) = INT_ZEXT EDX
    (unique,0x45d0,8) = INT_ZEXT EDX
    (unique,0x45e0,8) = INT_LEFT (unique,0x45d0,8), 0x20
    (unique,0x45f0,8) = INT_ZEXT EAX
    (unique,0x4610,8) = INT_OR (unique,0x45e0,8), (unique,0x45f0,8)
    (unique,0x4630,8) = INT_DIV (unique,0x4610,8), (unique,0x45c0,8)
    EAX = SUBPIECE (unique,0x4630,8), 0x0
    (unique,0x4660,8) = INT_REM (unique,0x4610,8), (unique,0x45c0,8)
    EDX = SUBPIECE (unique,0x4660,8), 0x0
```


## R_ANAL_OP_TYPE_SHR   = 22,

Pattern:
* Has `INT_RIGHT`
* Input of `INT_RIGHT` is a register or comes from register appearing in assembly
* Output of `INT_RIGHT` is a register or comes from register appearing in assembly

Example:
```asm
0x00001208: SHR EAX,1
    (unique,0xf50,1) = COPY 0x1
    (unique,0x80c0,4) = INT_AND EAX, 0x1
    CF = INT_NOTEQUAL (unique,0x80c0,4), 0x0
    OF = COPY 0x0
    EAX = INT_RIGHT EAX, 0x1
    SF = INT_SLESS EAX, 0x0
    ZF = INT_EQUAL EAX, 0x0
    (unique,0x1b20,1) = POPCOUNT EAX
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_SHL   = 23,

Pattern:
* Has `INT_LEFT`
* Input of `INT_RIGHT` is a register or comes from register appearing in assembly
* Output of `INT_RIGHT` is a register or comes from register appearing in assembly

Example:
```asm
0x0000120a: SHL EAX,1
    (unique,0xf50,1) = COPY 0x1
    CF = INT_SLESS EAX, 0x0
    EAX = INT_LEFT EAX, 0x1
    (unique,0x7b50,1) = INT_SLESS EAX, 0x0
    OF = INT_XOR CF, (unique,0x7b50,1)
    SF = INT_SLESS EAX, 0x0
    ZF = INT_EQUAL EAX, 0x0
    (unique,0x1b20,1) = POPCOUNT EAX
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_SAL   = 24,

Attention:
Isn't this equal to `R_ANAL_OP_TYPE_SHL`?

## R_ANAL_OP_TYPE_SAR   = 25,

Pattern:
* Has `INT_SRIGHT`
* Input of `INT_SRIGHT` is a register or comes from register appearing in assembly
* Output of `INT_SRIGHT` is a register or comes from register appearing in assembly

Example:
```asm
0x08048430: SAR EBX,0x2
    (unique,0x7900,4) = INT_AND 0x2, 0x1f
    (unique,0x7910,4) = COPY EBX
    EBX = INT_SRIGHT EBX, (unique,0x7900,4)
    (unique,0x2060,1) = INT_NOTEQUAL (unique,0x7900,4), 0x0
    (unique,0x2070,4) = INT_SUB (unique,0x7900,4), 0x1
    (unique,0x2080,4) = INT_SRIGHT (unique,0x7910,4), (unique,0x2070,4)
    (unique,0x2090,4) = INT_AND (unique,0x2080,4), 0x1
    (unique,0x20b0,1) = INT_NOTEQUAL (unique,0x2090,4), 0x0
    (unique,0x20c0,1) = BOOL_NEGATE (unique,0x2060,1)
    (unique,0x20d0,1) = INT_AND (unique,0x20c0,1), CF
    (unique,0x20e0,1) = INT_AND (unique,0x2060,1), (unique,0x20b0,1)
    CF = INT_OR (unique,0x20d0,1), (unique,0x20e0,1)
    (unique,0x2110,1) = INT_EQUAL (unique,0x7900,4), 0x1
    (unique,0x2120,1) = BOOL_NEGATE (unique,0x2110,1)
    OF = INT_AND (unique,0x2120,1), OF
    (unique,0x1b50,1) = INT_NOTEQUAL (unique,0x7900,4), 0x0
    (unique,0x1b70,1) = INT_SLESS EBX, 0x0
    (unique,0x1b80,1) = BOOL_NEGATE (unique,0x1b50,1)
    (unique,0x1b90,1) = INT_AND (unique,0x1b80,1), SF
    (unique,0x1ba0,1) = INT_AND (unique,0x1b50,1), (unique,0x1b70,1)
    SF = INT_OR (unique,0x1b90,1), (unique,0x1ba0,1)
    (unique,0x1bd0,1) = INT_EQUAL EBX, 0x0
    (unique,0x1be0,1) = BOOL_NEGATE (unique,0x1b50,1)
    (unique,0x1bf0,1) = INT_AND (unique,0x1be0,1), ZF
    (unique,0x1c00,1) = INT_AND (unique,0x1b50,1), (unique,0x1bd0,1)
    ZF = INT_OR (unique,0x1bf0,1), (unique,0x1c00,1)
    (unique,0x1c20,1) = POPCOUNT EBX
    PF = INT_AND (unique,0x1c20,1), 0x1
```

## R_ANAL_OP_TYPE_OR    = 26,

Pattern:
* Has `INT_OR`
* Input of `INT_OR` is a register or comes from register appearing in assembly
* Output of `INT_OR` is a register or comes from register appearing in assembly

Example:
```asm
0x00001208: OR ESP,dword ptr [0xfffffff0]
    CF = COPY 0x0
    OF = COPY 0x0
    ESP = INT_OR ESP, dword_ptr(0xfffffff0)
    SF = INT_SLESS ESP, 0x0
    ZF = INT_EQUAL ESP, 0x0
    (unique,0x1b20,1) = POPCOUNT ESP
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_AND   = 27,

Pattern:
* Has `INT_AND`
* Input of `INT_AND` is a register or comes from register appearing in assembly
* Output of `INT_AND` is a register or comes from register appearing in assembly

Example:
```asm
0x080483d5: AND ESP,0xfffffff0
    CF = COPY 0x0
    OF = COPY 0x0
    ESP = INT_AND ESP, 0xfffffff0
    SF = INT_SLESS ESP, 0x0
    ZF = INT_EQUAL ESP, 0x0
    (unique,0x1b20,1) = POPCOUNT ESP
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_XOR   = 28,

Pattern:
* Has `INT_XOR`
* Input of `INT_XOR` is a register or comes from register appearing in assembly
* Output of `INT_XOR` is a register or comes from register appearing in assembly

Example:
```asm
0x00001203: XOR EAX,EAX
    CF = COPY 0x0
    OF = COPY 0x0
    EAX = INT_XOR EAX, EAX
    SF = INT_SLESS EAX, 0x0
    ZF = INT_EQUAL EAX, 0x0
    (unique,0x1b20,1) = POPCOUNT EAX
    PF = INT_AND (unique,0x1b20,1), 0x1
```

## R_ANAL_OP_TYPE_NOR   = 29,

Pattern:
* Has `INT_OR` and `INT_NEGATE`
* Input of `INT_OR` is a register or comes from register appearing in assembly
* Output of `INT_OR` is a unique varnodedata
* Output of `INT_OR` is input of `INT_NEGATE`
* Output of `INT_NEGATE` is a register or comes from register appearing in assembly

Example:
```asm
0x00005758: nor r27,r27,r27
    (unique,0x69e0,4) = INT_OR r27, r27
    r27 = INT_NEGATE (unique,0x69e0,4)
```

## R_ANAL_OP_TYPE_NOT   = 30,

Pattern:
* has `INT_NEGATE`
* Input of `INT_XOR` is a register or comes from register appearing in assembly
* Output of `INT_XOR` is a register or comes from register appearing in assembly

Attention:
I guess its accuracy will be low.

Example:
```asm
0x00001208: NOT EAX
    EAX = INT_NEGATE EAX
```

## R_ANAL_OP_TYPE_LEA   = 33, /* TODO add ulea */

Attention:
Just ignore this.

Example:
```asm
0x00001129: LEA ESI,[ESI]
    ESI = COPY ESI
```

## R_ANAL_OP_TYPE_LEAVE = 34,

Attention:
Just ignore this.

Example:
```asm
0x000011ca: LEAVE 
    ESP = COPY EBP
    EBP = LOAD [ESP]
    ESP = INT_ADD ESP, 0x4
```

## R_ANAL_OP_TYPE_ROR   = 35,

Attention:
I don't think we can differ this from right shift itself.

Example:
```asm
0x00001210: ROR EAX,1
    (unique,0xf50,1) = COPY 0x1
    (unique,0x7500,4) = INT_AND EAX, 0x1
    CF = INT_NOTEQUAL (unique,0x7500,4), 0x0
    (unique,0x7520,4) = INT_RIGHT EAX, 0x1
    (unique,0x7530,4) = INT_ZEXT CF
    (unique,0x7540,4) = INT_LEFT (unique,0x7530,4), 0x1f
    EAX = INT_OR (unique,0x7520,4), (unique,0x7540,4)
    (unique,0x7560,4) = INT_AND EAX, 0x40000000
    (unique,0x7570,1) = INT_NOTEQUAL (unique,0x7560,4), 0x0
    (unique,0x7580,1) = INT_SLESS EAX, 0x0
    OF = INT_XOR (unique,0x7570,1), (unique,0x7580,1)

0x00001210: ROR EAX,0x5
    (unique,0x7610,4) = INT_AND 0x5, 0x1f
    (unique,0x7620,4) = INT_RIGHT EAX, (unique,0x7610,4)
    (unique,0x7630,4) = INT_SUB 0x20, (unique,0x7610,4)
    (unique,0x7640,4) = INT_LEFT EAX, (unique,0x7630,4)
    EAX = INT_OR (unique,0x7620,4), (unique,0x7640,4)
    (unique,0x1e00,1) = INT_NOTEQUAL (unique,0x7610,4), 0x0
    (unique,0x1e20,1) = INT_SLESS EAX, 0x0
    (unique,0x1e30,1) = BOOL_NEGATE (unique,0x1e00,1)
    (unique,0x1e40,1) = INT_AND (unique,0x1e30,1), CF
    (unique,0x1e50,1) = INT_AND (unique,0x1e00,1), (unique,0x1e20,1)
    CF = INT_OR (unique,0x1e40,1), (unique,0x1e50,1)
    (unique,0x1e80,1) = INT_EQUAL (unique,0x7610,4), 0x1
    (unique,0x1e90,1) = INT_SLESS EAX, 0x0
    (unique,0x1ea0,4) = INT_LEFT EAX, 0x1
    (unique,0x1eb0,1) = INT_SLESS (unique,0x1ea0,4), 0x0
    (unique,0x1ed0,1) = INT_XOR (unique,0x1e90,1), (unique,0x1eb0,1)
    (unique,0x1ee0,1) = BOOL_NEGATE (unique,0x1e80,1)
    (unique,0x1ef0,1) = INT_AND (unique,0x1ee0,1), OF
    (unique,0x1f00,1) = INT_AND (unique,0x1e80,1), (unique,0x1ed0,1)
    OF = INT_OR (unique,0x1ef0,1), (unique,0x1f00,1)
```

## R_ANAL_OP_TYPE_ROL   = 36,

Attention:
Like above.

Example:
```asm
0x00001213: ROL EAX,0x5
    (unique,0x7210,4) = INT_AND 0x5, 0x1f
    (unique,0x7220,4) = INT_LEFT EAX, (unique,0x7210,4)
    (unique,0x7230,4) = INT_SUB 0x20, (unique,0x7210,4)
    (unique,0x7240,4) = INT_RIGHT EAX, (unique,0x7230,4)
    EAX = INT_OR (unique,0x7220,4), (unique,0x7240,4)
    (unique,0x1ce0,1) = INT_NOTEQUAL (unique,0x7210,4), 0x0
    (unique,0x1cf0,4) = INT_AND EAX, 0x1
    (unique,0x1d10,1) = INT_NOTEQUAL (unique,0x1cf0,4), 0x0
    (unique,0x1d20,1) = BOOL_NEGATE (unique,0x1ce0,1)
    (unique,0x1d30,1) = INT_AND (unique,0x1d20,1), CF
    (unique,0x1d40,1) = INT_AND (unique,0x1ce0,1), (unique,0x1d10,1)
    CF = INT_OR (unique,0x1d30,1), (unique,0x1d40,1)
    (unique,0x1d70,1) = INT_EQUAL (unique,0x7210,4), 0x1
    (unique,0x1d80,1) = INT_SLESS EAX, 0x0
    (unique,0x1da0,1) = INT_XOR CF, (unique,0x1d80,1)
    (unique,0x1db0,1) = BOOL_NEGATE (unique,0x1d70,1)
    (unique,0x1dc0,1) = INT_AND (unique,0x1db0,1), OF
    (unique,0x1dd0,1) = INT_AND (unique,0x1d70,1), (unique,0x1da0,1)
    OF = INT_OR (unique,0x1dc0,1), (unique,0x1dd0,1)
```

## R_ANAL_OP_TYPE_XCHG  = 37,

* Has `COPY`
* Classic exchange procedure of two variables with a temporary variable

Example:
```asm
0x00001216: XCHG EAX,EBX
    (unique,0x84b0,4) = COPY EAX
    EAX = COPY EBX
    EBX = COPY (unique,0x84b0,4)
```

## R_ANAL_OP_TYPE_MOD   = 38,

Pattern:
* Has `INT_SREM` or `INT_REM`

Attention:
* Unluckily, assembly need improvement. P-code result is right, `v1` and `v0` is dalvik's local variables.

Example:
```asm
0x0000021e: rem_int_2addr 0x1004,0x1000
    v1 = INT_SREM v1, v0
```

## R_ANAL_OP_TYPE_SWITCH = 39,

Example:
```asm
0x0000021e: packed_switch 0x1004,0x55
    (unique,0x330,4) = INT_MULT 0x55, 0x2
    (unique,0x340,4) = INT_ADD 0x21e, (unique,0x330,4)
    (unique,0x350,4) = INT_ADD (unique,0x340,4), 0x2
    (unique,0x370,2) = LOAD [(unique,0x350,4)]
    (unique,0x390,4) = INT_ZEXT (unique,0x370,2)
    (unique,0x3a0,4) = INT_ADD 0x21e, (unique,0x330,4)
    (unique,0x3b0,4) = INT_ADD (unique,0x3a0,4), 0x2
    (unique,0x3c0,4) = INT_ADD (unique,0x3b0,4), 0x2
    (unique,0x3e0,4) = LOAD [(unique,0x3c0,4)]
    (unique,0x3f0,1) = INT_LESS v1, (unique,0x3e0,4)
    CBRANCH dword_ptr(0x00000224), (unique,0x3f0,1)
    (unique,0x400,4) = INT_ADD (unique,0x3e0,4), (unique,0x390,4)
    (unique,0x410,1) = INT_LESSEQUAL (unique,0x400,4), v1
    CBRANCH dword_ptr(0x00000224), (unique,0x410,1)
    (unique,0x420,4) = INT_ADD 0x21e, (unique,0x330,4)
    (unique,0x430,4) = INT_ADD (unique,0x420,4), 0x2
    (unique,0x440,4) = INT_ADD (unique,0x430,4), 0x2
    (unique,0x460,4) = INT_ADD (unique,0x440,4), 0x4
    (unique,0x480,4) = INT_SUB v1, (unique,0x3e0,4)
    (unique,0x490,4) = INT_MULT (unique,0x480,4), 0x4
    (unique,0x4a0,4) = INT_ADD (unique,0x460,4), (unique,0x490,4)
    (unique,0x4c0,4) = LOAD [(unique,0x4a0,4)]
    (unique,0x4d0,4) = INT_MULT (unique,0x4c0,4), 0x2
    (unique,0x4f0,4) = INT_ADD 0x21e, (unique,0x4d0,4)
    BRANCHIND (unique,0x4f0,4)
```

## R_ANAL_OP_TYPE_CASE = 40,
No related pcode

## R_ANAL_OP_TYPE_LENGTH = 41,

Example:
```asm
0x00000234: array_length 0x1008,0x1008
    v2 = CPOOLREF v2, 0x0, 0x6
```

## R_ANAL_OP_TYPE_CAST = 42,

Pattern:
* Has `CAST`

## R_ANAL_OP_TYPE_NEW = 43,

Pattern:
* Has `NEW`

## R_ANAL_OP_TYPE_ABS = 44,

Pattern:
* Has `FLOAT_ABS`

Example:
```asm
0x00001208: FABS 
    ST0 = FLOAT_ABS ST0
```

## R_ANAL_OP_TYPE_CPL = 45,	/* complement */

Attention:
Just merge this situation into `R_ANAL_OP_TYPE_NOT`

Example:
```asm
0x00300002      comf 0x18, 1, 0
            unique(11e0, 1) = INT_NEGATE 18
            18 = COPY unique(11e0, 1)
            N = INT_SLESS unique(11e0, 1) 0
            Z = INT_EQUAL unique(11e0, 1) 0
```

## R_ANAL_OP_TYPE_CRYPTO = 46,

Attention:
Just ignore this. Merge this into syscall

## R_ANAL_OP_TYPE_SYNC = 47,

Attention:
Just ignore this. Merge this into syscall

wargio:
> Fantastic overview!
> 
> > > It's recognize as `COMPUTED_CALL` due to `CALLIND`. Do I need to specify this is `R_ANAL_OP_TYPE_SWI`?
> > 
> > 
> > No, i do not think is needed.
> 
> I think it is needes because syscalls are very different instructions from calls. Since you already identified that they always have this `CALLOTHER`, do you think it would be hard to differenciate them?
> 
I do agree but my issue is more what for sleigh converts that interrupt to.

FXTi:
@wargio About `CALLOTHER`, it's actually SLEIGH's "custom op" like ESIL.
SLEIGH read userop from config. So it's basically any op which can not fit in existing P-code ecosystem, like CRYPTO or SYNC type in r2.
I think it's acceptable to classify them as syscall.
It's like extra function provided by system through interrupt.

FXTi:
Now, all new features are complete. You can begin review, and I will fix bugs. I guess there will be tons of bugs...

Btw, this need even more tests than asm_ghidra. So if you have any idea to generate tests case fast, plz leave comments!

@thestr4ng3r @XVilka @wargio

FXTi:
Note: SleighInstruction cache will consume much memory, I need to limit its size.

src/SleighAsm.cpp
```diff
@@ -17,30 +19,161 @@ void SleighAsm::init(RAsm *a)
 	if(!io)
 		throw LowlevelError("Can't get RIO from RBin");
 
+	initInner(io, a->cpu);
+}
+
+void SleighAsm::init(RAnal *a)
+{
+	if(description.empty())
```
thestr4ng3r, line 27: Too much code duplication here from `SleighAsm::init(RAsm *a)`. It will be better to merge in one function `SleighAsm::init(const char *sleigh_id, RIO *io, RConfig *cfg)`.

src/SleighAsm.cpp
```diff
@@ -17,30 +19,161 @@ void SleighAsm::init(RAsm *a)
...
+
+	const List &list(el->getChildren());
+	List::const_iterator iter;
+
+	for(iter = list.begin(); iter != list.end(); iter++)
+	{
+		const string &elname((*iter)->getName());
+		if(elname == "stackpointer")
+			sp_name = (*iter)->getAttributeValue("register");
+		if(elname == "default_proto")
```
thestr4ng3r, line 136:
```suggestion
		else if(elname == "default_proto")
```

src/SleighAsm.cpp
```diff
@@ -322,7 +474,13 @@ int SleighAsm::disassemble(RAsmOp *op, unsigned long long offset)
 	try
 	{
 		length = trans.printAssembly(assem, addr);
-		r_strbuf_set(&op->buf_asm, assem.str);
+		// r_strbuf_set(&op->buf_asm, assem.str);
+		auto *ins = trans.getInstruction(addr);
+		stringstream ss;
+		ss << assem.str << " " << ins->printFlowType(ins->getFlowType());
```
thestr4ng3r, line 480: Is this debugging code? If you want to keep it, you can wrap it in `#if 0` to disable

src/SleighAsm.h
```diff
@@ -6,67 +6,274 @@
...
+	~AssemblySlg()
+	{
+		if(str)
+			free(str);
+	}
+};
+
+struct PcodeOperand
+{
+	PcodeOperand(): PcodeOperand(0x7fffffff, 0x7fffffff) {}
```
thestr4ng3r, line 51: Is this constructor used anywhere? Where do the `0x7fffffff` come from?

src/SleighAsm.h
```diff
@@ -6,67 +6,274 @@
...
+			pcodes.pop_back();
+		}
+	}
+};
+
+struct R2Reg
+{
+	std::string name;
+	size_t size;
+	size_t offset;
```
thestr4ng3r, line 235: This and `size` should be `ut64` I think.

src/SleighInstruction.h
```diff
@@ -0,0 +1,452 @@
...
+ * like InstructionContext and move necessary status and API to that.
+ */
+
+class R2Sleigh;
+class R2DisassemblyCache;
+class SubParserWalker;
+class SleighParserContext;
+class ExportHelper
+{
+	// Please keep all structures in this class sync with origin!!!
```
thestr4ng3r, line 33: These structures have to be removed before merge. You can send a pr to https://github.com/thestr4ng3r/ghidra to expose things.

src/SleighInstruction.h
```diff
@@ -0,0 +1,452 @@
...
+		return ((ExportHelper::ParserWalker *)xxx)->const_context;
+	}
+	static ::ContextCache *getCache(R2Sleigh *xxx) { return ((ExportHelper::Sleigh *)xxx)->cache; }
+	static ::ConstructState **getBasestate(SleighParserContext *xxx)
+	{
+		return &((ExportHelper::ParserContext *)xxx)->base_state;
+	}
+};
+
+typedef enum
```
thestr4ng3r, line 98: Better to use C++11 `enum class`, especially since you are already using the values scoped with `FlowType::`.

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,2405 @@
...
+	else // Label each instruction based on a series of P-codes.
+		anal_type(a, anal_op, pcode_slg, assem);
+
+	return anal_op->size;
+}
+
+static char *get_reg_profile(RAnal *anal)
+{
+	if(!strcmp(anal->cpu, "x86"))
+		return nullptr;
```
thestr4ng3r, line 1735: What's the reason for this?

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,2405 @@
...
+{
+	bool ret = false;
+	long double s;
+	char *src = r_anal_esil_pop(esil);
+	if(src && esil_get_parm_float(esil, src, &s))
+		ret = r_anal_esil_pushnum(esil, isnan(s));
+	else
+		throw LowlevelError("sleigh_esil_is_nan: invalid parameters");
+
+	free(src);
```
thestr4ng3r, line 1923: These frees on memory malloced in r2 can become problematic depending on the C++ compiler/lib. Better use `r_mem_free()` for those.

thestr4ng3r (changes requested):

thestr4ng3r:
Seems to crash currently, depending on the used sleigh id:
20200824_22h41m28s_grim

FXTi:
@thestr4ng3r This is quite interesting: if you specify `asm.cpu` first, everything will be fine.

FXTi:
21~M / 1024 instructions

src/SleighAsm.cpp
FXTi: Yes, will do.

src/SleighAsm.h
FXTi: This is constructor, used to specify it's uninitialized.
I will remove this.

src/SleighAsm.h
FXTi: Yes

src/anal_ghidra.cpp
FXTi: To avoid unnecessary initialization when radare just try to guess what arch current binary is.

FXTi:
Still problem in FP to mem, mem to FP
```
[0x00000000]> e asm.arch=x86
[0x00000000]> e asm.bits=32
[0x00000000]> pdga
[0x00000000]> "ae 7,0x0,=[8]"
[0x00000000]> "ae 0x0,[8]"
0x7
[0x00000000]> "ae 7,I2F,0x0,=[8]"
[0x00000000]> "ae 0x0,[8]"
0x401c000000000000
[0x00000000]> "ae 7,I2F,512,=[8]"
[0x00000000]> "ae 512,[8]"
0x401c000000000000
[0x00000000]> "ae 7,I2F,512+8,=[8]"
[0x00000000]> "ae 512+16,[8]"
0x401c000000000000
[0x00000000]> "ae 512+32,[8]"
0x401c000000000000
[0x00000000]> 
```

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,3343 @@
...
+	return -1;
+}
+
+static void append_hardcoded_regs(std::stringstream &buf, const std::string &arch, bool little,
+                                  int bits)
+{
+	if(arch.size() < 3)
+		throw LowlevelError("append_hardcoded_regs: Unexpected arch name.");
+
+	switch(arch[0] | arch[1] << 8 | arch[2] << 16)
```
thestr4ng3r, line 1987: I don't think you have to optimize the switch on the string this hard. As the nicest solution I would recommend to do something like here:
https://github.com/radareorg/r2ghidra-dec/blob/8e576eeadc211de4ac8d8c759cc368fa48cdfa99/src/ArchMap.cpp#L67-L79
Maybe you can even re-use the Mapper classes there if you move them to the header and replace `RCore *` here by a template parameter:
https://github.com/radareorg/r2ghidra-dec/blob/8e576eeadc211de4ac8d8c759cc368fa48cdfa99/src/ArchMap.cpp#L10-L14

But alternatively, a simple if/else would work too if you don't want to over-engineer it too much.

src/anal_ghidra.cpp
thestr4ng3r: So "x86" in this case is because it's the default arch that r2 sets before knowing it from the binary? This could be problematic because the default arch in r2 is dependent on the host architecture, i.e. if you work on a raspberry pi or something, it would be "arm".
In any case, I think this needs at least a comment so it is clear why this check is there.

src/core_ghidra.cpp
```diff
@@ -100,6 +100,7 @@ static void PrintUsage(const RCore *const core)
 		CMD_PREFIX"s",  "", "# Display loaded Sleigh Languages",
 		CMD_PREFIX"ss", "", "# Display automatically matched Sleigh Language ID",
 		CMD_PREFIX"sd", " N", "# Disassemble N instructions with Sleigh and print pcode",
+		CMD_PREFIX"a", "", "# Enable RAsm and RAnal plugins driven by SLEIGH from Ghidra",
```
thestr4ng3r, line 103:
```suggestion
		CMD_PREFIX"a", "", "# Switch to RAsm and RAnal plugins driven by SLEIGH from Ghidra",
```
I think this is a bit more clear about what it does.

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,3343 @@
...
+	if(!esil || !esil->stack)
+		return false;
+
+	char *idx = r_anal_esil_pop(esil);
+	ut64 i;
+	int ret = false;
+
+	if(R_ANAL_ESIL_PARM_REG == r_anal_esil_get_parm_type(esil, idx))
+	{
+		throw LowlevelError("sleigh_esil_consts_pick: argument is consts only");
```
thestr4ng3r, line 2303: These throws in the esil handlers can be problematic since they crash the entire process if something is wrong in an esil string, which can theoretically always happen because the user can also supply esil. I think a simple print would be enough for now.

thestr4ng3r (changes requested):
I think we should have some specific testing for the whole pattern-matching. What I would imagine are tests that write specific instructions (without a binary) and then do `ao` to print the instruction info.
Something like this:
```
e asm.arch=r2ghidra
e asm.cpu=x86:...
wx 31ed
ao
?e --
wx ...
ao
...
```
Normally you would write one r2r test for each of these ops, but since sleigh always takes so long to initialize I think one test with `?e --` between the instructions should be ok.

If you add one such check for each of the instruction types in your overview here: https://github.com/radareorg/r2ghidra-dec/pull/120#issuecomment-673083656 (excluding the more exotic stuff we ignored), then I think we might be good to merge.

src/anal_ghidra.cpp
thestr4ng3r: Also, it should not fail completely if the arch is not known. You should still be able to use the plugin at least with unrecognized BP and SN.

FXTi: I will save this for next pr. Since current map is got from existing RAnal plugin source code, I think their result is not good enough. There must be something missing. I prepare to redo this in the future.

XVilka (approved):
Lets merge this, further improvements can be done as a separate pull requests.

src/SleighInstruction.cpp
```diff
@@ -0,0 +1,462 @@
...
+	int offset = getLength();
+	int bytecount = 0;
+	do
+	{
+		Address off_addr = baseaddr + offset;
+		SleighInstruction ins(sleigh, off_addr);
+		int len = ins.getLength();
+		offset += len;
+		bytecount += len;
+	} while(bytecount < delaySlotByteCnt);
```
wargio, line 460: Probably a check with len < 0 should be added?

src/anal_ghidra.cpp
wargio: I agree with both. Yes no need to optimize that and no need to make a new commit to address this now.

src/anal_ghidra.cpp
wargio: Yes please add a comment.

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,3553 @@
...
+				memcpy(&vd, regset->arena->bytes + off, sizeof(double));
+				ret = vd;
+			}
+			break;
+		case 80:
+		case 96:
+		case 128:
+			if(regset->arena->size - off - 1 >= 0)
+			{
+				memcpy(&ret, regset->arena->bytes + off, sizeof(long double));
```
wargio, line 2173: I do not think using `long double` is a good idea.

wargio: Not all compilers supports 64+ bits regs. I suggest to limit this to x64 via ifdef endif

src/anal_ghidra.cpp
```diff
@@ -0,0 +1,3553 @@
...
+			}
+			break;
+		default:
+			eprintf("esil_get_double: Bit size not supported.\n");
+			return 0.0f;
+	}
+	return ret;
+}
+
+static bool esil_set_double(RReg *reg, RRegItem *item, long double value)
```
wargio, line 2183: Same this

wargio (changes requested):

wargio (approved):

wargio (approved):

src/SleighInstruction.cpp
FXTi: Agree

thestr4ng3r (approved):

PR #121, opened 2020-06-26
Implemented offset annotation for comments

NirmalManoj:
### **Detailed description**

Earlier we didn't have offset annotation (`R_CODE_ANNOTATION_TYPE_OFFSET`) for comments. Because of this, we have inconsistent behavior while clicking on a comment to edit in Cutter. This PR solves this problem by implementing offset annotations for comments.

The following GIFs show the previous behavior and improved behavior.
[Note: The following GIFs don't use the current master of Cutter. The same behavior can be noticed from the master also. So this PR can be merged after testing]

**Before having offset annotation**
BeforeCommentOffset
**After implementing offset annotation**
AfterCommentOffset_1

...

**Test plan**
- Test with the current master of Cutter
- Test if the changes are working as expected
- Make sure nothing else is affected by this change

...

**Closing issues**

...

### DIFF
--- a/src/CodeXMLParse.cpp
+++ b/src/CodeXMLParse.cpp
@@ -47,6 +47,20 @@ void AnnotateOpref(ANNOTATOR_PARAMS)
 	annotation.offset.offset = op->getAddr().getOffset();
 }
 
+void AnnotateCommentOffset(ANNOTATOR_PARAMS){
+	pugi::xml_attribute attr = node.attribute("off");
+	if(attr.empty())
+		return;
+	unsigned long long off = attr.as_ullong(ULLONG_MAX);
+	if(off == ULLONG_MAX)
+		return;
+	out->emplace_back();
+	auto &annotation = out->back();
+	annotation = {};
+	annotation.type = R_CODE_ANNOTATION_TYPE_OFFSET;
+	annotation.offset.offset = off;
+}
+
 /**
  * Translate Ghidra's color annotations, which are essentially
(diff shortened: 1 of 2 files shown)


karliss:
Broken test. It would also be good to test the offset for r2 comment instead of ghidra generated warning.

NirmalManoj:
@karliss Tests will likely work if we stop giving offset annotations for ghidra's warnings.

karliss:
>Tests will likely work if we stop giving offset annotations for ghidra's warnings.

Yes, but those offsets aren't exactly wrong. I think it's fine updating the test in this case. Is there even an easy way to distinguish source of comment?

NirmalManoj:
> It would also be good to test the offset for r2 comment instead of ghidra generated warning.

> Is there even an easy way to distinguish source of comment?

I was looking to find a way to do this. Other than modifying some parts of the real ghidra that generate warnings, I couldn't find any easy way to do this. Even that will be sort of like a hack.

If we decide to leave it like this, this weird behavior will exist while clicking on a warning to add a comment. But as you said, I think it's okay as long as it won't affect users from adding/editing their own comments.

NirmalManoj:
Is Jessie CI down? In PR `#120`, last commit was three days ago and Jessie's status is still _Expected_.

karliss:
@NirmalManoj I am still waiting for the test which checks the offset of user created comment.

NirmalManoj:
@karliss I have created a test that checks the offset of user-created comments.

PR #123, opened 2020-07-03
Function name annotation annotator implemented

NirmalManoj:
**Detailed description**
This PR implements an annotator that will annotate function names with a new type of annotation (`R_CODE_ANNOTATION_TYPE_FUNCTION_NAME`) introduced by [PR #17204](https://github.com/radareorg/radare2/pull/17204) in radare2. This will be required for implementing some features in the decompiler widget in Cutter.

I have also implemented [offset annotation](https://github.com/radareorg/r2ghidra-dec/pull/123#discussion_r451417915) for the currently decompiled function's name in this PR.

...

**Test plan**
1. Fetch [PR #17204](https://github.com/radareorg/radare2/pull/17204) in radare2 and compile it before compiling this.
2. Make sure the output is working as expected from the JSON output (check r2 PR first for making sure JSON output is implemented correctly for function names).
3. Check code and other things that should be checked.

...

**Closing issues**

...


karliss:
Same as with r2 changes, and cutter should better go into decompiler-refactoring branch until the whole changes are more complete and API is stabilized.

karliss:
Since json output should contains this you can easily create a test for this.

karliss:
What kind of actions are you planning to implement using this annotation?

NirmalManoj:
@karliss The next action I am going to add is an action to rename functions. This annotation is required for it. Other than this, I don't have anything else in mind where we will need this necessarily. But I believe it will be useful in some other place.

karliss:
Take a look at help for  `afn`  command used to rename function. It is `afn[?] name [addr] rename name for function at address (change flag too)".  It would be more reliable to use address of the function directly instead of the name. While the name can be evaluated to obtain address it's messy and you need the correct name. There is function name, real name, flag name, demangled name.  The address of function referenced at position of code would also have other uses.

src/CodeXMLParse.cpp
```diff
@@ -46,6 +46,15 @@ void AnnotateOpref(ANNOTATOR_PARAMS)
 	annotation.type = R_CODE_ANNOTATION_TYPE_OFFSET;
 	annotation.offset.offset = op->getAddr().getOffset();
 }
+void AnnotateFunctionName(ANNOTATOR_PARAMS){
+	auto func_name = node.child_value();
```
NirmalManoj, line 50: (not related to the code here, making this comment to start a nested discussion)
> It would be more reliable to use address of the function directly instead of the name. While the name can be evaluated to obtain address it's messy and you need the correct name. There is function name, real name, flag name, demangled name. The address of function referenced at position of code would also have other uses.

I didn't think that renaming using the name would get messy. But initially, I wanted to have the address in the function name annotations, just so that it might become useful later. But unfortunately, we don't have this information in XML. So I am not sure how I could do it @karliss  @thestr4ng3r

NirmalManoj: https://github.com/radareorg/radare2/blob/master/libr/include/r_anal.h#L277-L280
Does this name always refer to the 'function name' and not real name/flag name/demangled name.

karliss: The address can be clearly seen in the xml.

NirmalManoj: @karliss I am unable to find it till now. We have the address of the current function in the XML(and I knew we could get that).   But how can I find the addresses of all the functions in the decompiled code from the XML?

thestr4ng3r: @NirmalManoj Can you post an example xml here for reference?

NirmalManoj:
[global_variable.xml.txt](https://github.com/radareorg/r2ghidra-dec/files/4874962/global_variable.xml.txt)
@thestr4ng3r Couldn't upload this file under that comment. After Karliss's comment, I did notice that the offsets are present under varnodes(but they are not directly connected by a reference), but I want help in parsing it.

src/CodeXMLParse.cpp
karliss: I had a call with Nirmal looking at the xml, Ghidra C++ API and Opref annotator as example for accessing Ghidra C++ API. He should be good for a while.

NirmalManoj: @karliss @thestr4ng3r  I was trying this all day and it seems I need more help. This is what I am trying finally (not working code)
```
        PcodeOp *op = opit->second;
	uintb temp_off = 0;
	for(int i = 0; i < op->numInput(); i++){
		const Varnode *vn = op->getIn(i);
		if (vn->getSpace()->getType() == IPTR_FSPEC){
			temp_off = vn->getOffset();
			// temp_off++;
		}
	}
	annotation.function_name.offset = temp_off;
	out->push_back(annotation);
```
It does identify functions correctly with `IPTR_FSPEC`, but `vn->getOffset()` is not returning the correct offset. I want to find a way to get the offset from the `Varnode` and I believe there is a way to do that from `vn`. But I couldn't figure that out.

karliss: a) How did you came to conclusion that IPTR_FSPEC is the correct address space kind?
b) did you to dump all the varnodes you get with their properties: type, whatever offset returns, the id displayed in xml, anything else potentially interesting by printing it to terminal?
c) The offsets in xml looked reasonable, take a look a t how xml is printed

NirmalManoj: a) In `space.hh` in Ghidra,
```
/// \brief Fundemental address space types
///
/// Every address space must be one of the following core types
enum spacetype {
  IPTR_CONSTANT = 0,	       ///< Special space to represent constants
  IPTR_PROCESSOR = 1,	       ///< Normal spaces modelled by processor
  IPTR_SPACEBASE = 2,	       ///< addresses = offsets off of base register
  IPTR_INTERNAL = 3,	       ///< Internally managed temporary space
  IPTR_FSPEC = 4,	       ///< Special internal FuncCallSpecs reference
  IPTR_IOP = 5,                ///< Special internal PcodeOp reference
  IPTR_JOIN = 6		       ///< Special virtual space to represent split variables
};
```
From this and the code that I have read, I guessed `IPTR_FSPEC` will be the only `spacetype` for function calls. So I tried it.
c) I was trying to find a way to get the offsets without extracting it from the XML. I did check saveXml functions (but not ruling out the possibility that I won't get anything from it in future)
b) I haven't done all of it. I will try it now.

NirmalManoj: I used a debugger to see the offsets of all the `varnodes` inside `ctx->func.vbank.loc_tree`. I will attach the binary below. It has 25 nodes under `varnodes` tag in XML output and it displays the offsets of all those correctly except those that give us function offsets. 
( Attached image shows the location that I am talking about clearly)
Screenshot from 2020-07-07 18-48-29 Screenshot from 2020-07-07 18-53-32

It will be great if someone with experience of Ghidra API could help me in understanding it more. I have spent a lot of time in the past two days trying to figure this out. @karliss @thestr4ng3r

NirmalManoj:
[a.zip](https://github.com/radareorg/r2ghidra-dec/files/4885067/a.zip)
This is the binary mentioned in https://github.com/radareorg/r2ghidra-dec/pull/123#discussion_r450862118

src/CodeXMLParse.cpp
NirmalManoj: I think I have finally found a way to extract offsets reliably.

src/CodeXMLParse.cpp
```diff
@@ -46,8 +45,67 @@ void AnnotateOpref(ANNOTATOR_PARAMS)
...
+	{
+		if(ctx->func->getName() == func_name)
+		{
+			annotation.function_name.name = strdup(ctx->func->getName().c_str());
+			annotation.function_name.offset = ctx->func->getAddress().getOffset();
+			// Code below makes an offset annotation for the function name(for the currently decompiled function)
+			RCodeAnnotation offsetAnnotation = {};
+			offsetAnnotation.type = R_CODE_ANNOTATION_TYPE_OFFSET;
+			offsetAnnotation.offset.offset = annotation.function_name.offset;
+			out->push_back(offsetAnnotation);
```
NirmalManoj, line 66: With the currently pushed commits, I have also implemented offset annotation for the currently decompiled function's name. See gif
offset_annotation_for_function_name

src/CodeXMLParse.cpp
NirmalManoj: @karliss @thestr4ng3r This PR is ready for more review.

src/CodeXMLParse.cpp
```diff
@@ -46,17 +45,67 @@ void AnnotateOpref(ANNOTATOR_PARAMS)
...
+	auto opit = ctx->ops.find((uintm)opref);
+	if(opit == ctx->ops.end())
+	{
+		annotation.function_name.name = strdup("INVALID: Doesn't look like a function.");
+		annotation.function_name.offset = UINT64_MAX;
+		out->push_back(annotation);
+		return;	
+	}
+	PcodeOp *op = opit->second;
+	FuncCallSpecs *call_func_spec = ctx->func->getCallSpecs(const_cast<PcodeOp *>(op));
```
karliss, line 93: Why did you decide to do a const cast here?

NirmalManoj: `FuncCallSpecs *getCallSpecs(const PcodeOp *op) const;`
Because it accepts a constant PcodeOp as the argument. Now I think I should remove it.

karliss: Code should almost never contain contain const_casts. Valid uses cases for it are very rare. In this case you are  not even changing the type using the cast. You might want to read a bit more about how const works.

NirmalManoj: Thanks, I understood why it's useless in this case.

NirmalManoj: Removed const cast.

.travis.yml
```diff
@@ -24,13 +24,13 @@ env:
...
   - pip3 install meson
-  - git clone --depth 1 https://github.com/radareorg/radare2
+  - git clone --depth 1 --single-branch --branch decompiler-refactoring https://github.com/radareorg/radare2
   - cd radare2 && mkdir build && cd build
   - meson --prefix="$INSTALL_PREFIX" ..
   - ninja
   - ninja install
   - cd ../..
-  - git clone --depth 1 https://github.com/radareorg/cutter
+  - git clone --depth 1 --single-branch --branch decompiler-refactoring https://github.com/radareorg/radare2 https://github.com/radareorg/cutter
```
thestr4ng3r, line 33:
```suggestion
  - git clone --depth 1 --single-branch --branch decompiler-refactoring https://github.com/radareorg/cutter
```

src/CodeXMLParse.cpp
```diff
@@ -46,8 +45,67 @@ void AnnotateOpref(ANNOTATOR_PARAMS)
...
+			annotation.function_name.offset = ctx->func->getAddress().getOffset();
+			// Code below makes an offset annotation for the function name(for the currently decompiled function)
+			RCodeAnnotation offsetAnnotation = {};
+			offsetAnnotation.type = R_CODE_ANNOTATION_TYPE_OFFSET;
+			offsetAnnotation.offset.offset = annotation.function_name.offset;
+			out->push_back(offsetAnnotation);
+		}
+		else
+		{
+			annotation.function_name.name = strdup("INVALID: Doesn't look like a function.");
```
thestr4ng3r, line 70: Just return without adding an annotation in these cases instead of adding an invalid one. Same for the other "INVALID" ones.

src/CodeXMLParse.h
```diff
@@ -9,4 +9,5 @@ class Funcdata;
 
 R_API RAnnotatedCode *ParseCodeXML(Funcdata *func, const char *xml);
 
+
```
thestr4ng3r, line 12:
```suggestion
```

src/CodeXMLParse.cpp
```diff
@@ -1,7 +1,7 @@
 
 #include "CodeXMLParse.h"
 #include <r_util/r_annotated_code.h>
-
+#include <iostream>
```
thestr4ng3r, line 4: Is this needed?

src/CodeXMLParse.cpp
```diff
@@ -11,7 +11,6 @@
 #include <pugixml.hpp>
 #include <sstream>
 #include <string>
-
```
thestr4ng3r, line 14: Keep it please.

src/CodeXMLParse.cpp
```diff
@@ -183,6 +241,7 @@ static void ParseNode(pugi::xml_node node, ParseCodeXMLContext *ctx, std::ostrea
 R_API RAnnotatedCode *ParseCodeXML(Funcdata *func, const char *xml)
 {
 	pugi::xml_document doc;
+
```
thestr4ng3r, line 244:
```suggestion
```

.travis.yml
NirmalManoj: I will fix this.

src/CodeXMLParse.cpp
NirmalManoj: I will change this as suggested

src/CodeXMLParse.cpp
NirmalManoj: Not at all, I used it for printing something.

src/CodeXMLParse.cpp
NirmalManoj: I have made all the suggested changes.

NirmalManoj: I'm sorry, I misread your suggestion. I will change it now. All the other changes are as suggested.

NirmalManoj: Done. Changed as suggested.

NirmalManoj:
@thestr4ng3r @karliss I have fixed tests. Jessie is still failing because of reasons I don't understand. I think this PR is good to merge now.

thestr4ng3r:
Looks like jessie is using r2 master. But that shouldn't be a blocker.

PR #128, opened 2020-07-30
Annotator for local variables and function paramters

NirmalManoj:
**Detailed description**

This PR implements an annotator for annotating local variables and function parameters. The annotator uses the color attribute in the XML for classifying variables as local variables and function parameters. 

...

**Test plan**
1. Fetch [PR #17375](https://github.com/radareorg/radare2/pull/17375)  from radare2 and compile it.
2. Load any function (test functions with local variables and functions without local variables).
3. Pipe the output of `pdgj~{}` to a file and open it in an editor of your choice.
4. Because of the order of annotations, syntax highlight annotator with come right after local variable anntoation.
5. Search for `"syntax_highlight": "function_parameter"` and `"syntax_highlight": "local_variable"`. Make sure that you are seeing the exact local variable that the syntax highlight points in the annotation above it. For example, it should like this.
```    
    {
      "start": 551,
      "end": 558,
      "type": "function_parameter",
      "name": "arg_10h"
    },
    {
      "start": 551,
      "end": 558,
      "type": "syntax_highlight",
      "syntax_highlight": "function_parameter"
    }, 
```
6. Then search for `"type": "function_parameter"` and `"type": "local_variable"` and make sure that you can see the corresponding syntax highlight annotation right after that.

...

**Closing issues**

...

### DIFF
--- a/src/CodeXMLParse.cpp
+++ b/src/CodeXMLParse.cpp
@@ -197,7 +197,7 @@ void AnnotateVariable(ANNOTATOR_PARAMS)
 		AnnotateGlobalVariable(varnode, out);
 	else if (varnode->getHigh()->isConstant() && varnode->getHigh()->getType()->getMetatype() == TYPE_PTR) 
 		AnnotateConstantVariable(varnode, out);
-	else (!varnode->getHigh()->isPersist() && (varnode->getHigh()->getSymbol() != (Symbol *)0))
+	else if (!varnode->getHigh()->isPersist() && (varnode->getHigh()->getSymbol() != (Symbol *)0))
 		AnnotateLocalVariable(node, out);
 }
 
--- a/src/CodeXMLParse.cpp
+++ b/src/CodeXMLParse.cpp
@@ -176,7 +176,7 @@ void AnnotateLocalVariable(pugi::xml_node node, std::vector<RCodeAnnotation> *ou
 	annotation.variable.name = strdup(node.child_value());
 	if (color == "param")
 		annotation.type = R_CODE_ANNOTATION_TYPE_FUNCTION_PARAMETER;
-    else if (color == "var")
+	else if (color == "var")
 		annotation.type = R_CODE_ANNOTATION_TYPE_LOCAL_VARIABLE;
 	out->push_back(annotation);
 }
@@ -197,7 +197,7 @@ void AnnotateVariable(ANNOTATOR_PARAMS)
 		AnnotateGlobalVariable(varnode, out);
 	else if (varnode->getHigh()->isConstant() && varnode->getHigh()->getType()->getMetatype() == TYPE_PTR) 
 		AnnotateConstantVariable(varnode, out);
-	else
+	else (!varnode->getHigh()->isPersist() && (varnode->getHigh()->getSymbol() != (Symbol *)0))
 		AnnotateLocalVariable(node, out);
 }
 
--- a/src/CodeXMLParse.cpp
+++ b/src/CodeXMLParse.cpp
@@ -49,6 +49,39 @@ struct ParseCodeXMLContext
 			}
 			
 		}
+		ScopeLocal *mapLocal = func->getScopeLocal();
+		// SymbolNameTree nametree = mapLocal->nametree;
+		// SymbolNameTree multiEntrySet = mapLocal->beginMultiEntry();
+		set<Symbol *> num;
+		auto nil = mapLocal->beginMultiEntry();
+		auto nilpotent = *nil;
+		Symbol *noice = nilpotent;
+		// auto nilkopotent = *nilpotent;
+		for (auto star = mapLocal->beginMultiEntry(); star != mapLocal->endMultiEntry(); star++) {
+			num.insert(*star);
+		}
+		// SymbolNameTree::const_iterator iter = mapLocal->beginMultiEntry();
+		// SymbolNameTree::const_iterator enditer = mapLocal->endMultiEntry();
+		// for(;iter!=enditer;++iter) {
+		// 	// vector<Varnode *> mergeList;
+    	// 	Symbol *symbol = *iter;
+		// 	int ususi = 10;
+		// }
+		int xcfsdfss = 10;
+		SymbolNameTree nametree = mapLocal->getNameTree();
+		int kiliol=1;
+		// auto xkoi = mapLocal->begin();
+		
+		MapIterator iter = mapLocal->begin();
+		MapIterator enditer = mapLocal->end();
+		set<Symbol *> numTree;
+		for(;iter!=enditer;++iter) {
+			const SymbolEntry *entry = *iter;
+			if (entry->isPiece()) continue; // Don't do a partial entry
+			Symbol *sym = entry->getSymbol();
+			num.insert(sym);
+		}
+		int kcik = 123;
 	}
 };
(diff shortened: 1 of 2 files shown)


src/CodeXMLParse.cpp
```diff
@@ -162,11 +162,30 @@ void AnnotateConstantVariable(Varnode *varnode, std::vector<RCodeAnnotation> *ou
...
+	if (color == "")
+		return;
+	
+	RCodeAnnotation annotation = {};
+	annotation.variable.name = strdup(node.child_value());
+	if (color == "param")
+		annotation.type = R_CODE_ANNOTATION_TYPE_FUNCTION_PARAMETER;
+    else if (color == "var")
+		annotation.type = R_CODE_ANNOTATION_TYPE_LOCAL_VARIABLE;
+	out->push_back(annotation);
```
karliss, line 181: Please do not make decisions based on color.

src/CodeXMLParse.cpp
```diff
@@ -162,11 +162,30 @@ void AnnotateConstantVariable(Varnode *varnode, std::vector<RCodeAnnotation> *ou
...
+		return;
+	std::string color = attr.as_string();
+	if (color == "")
+		return;
+	
+	RCodeAnnotation annotation = {};
+	annotation.variable.name = strdup(node.child_value());
+	if (color == "param")
+		annotation.type = R_CODE_ANNOTATION_TYPE_FUNCTION_PARAMETER;
+    else if (color == "var")
```
thestr4ng3r, line 179:
```suggestion
	else if (color == "var")
```

thestr4ng3r:
As a hint, this is the decision that leads to the color:
https://github.com/thestr4ng3r/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc#L1560-L1563

src/CodeXMLParse.cpp
NirmalManoj: @karliss  @thestr4ng3r I had seen the decision that leads to color. For PR #124 , I have used info from here. https://github.com/thestr4ng3r/ghidra/blob/6c10f36f06468f866188cccf960c019779fb9028/Ghidra/Features/Decompiler/src/decompile/cpp/variable.cc#L452-L470

In there also, the same logic (if I didn't miss out anything) is used there to classify local variables and parameters. This is the reason why I used color to classify a variable as a parameter and local variable. Have you seen examples where color is mistakenly annotated?

src/CodeXMLParse.cpp
NirmalManoj: @karliss @thestr4ng3r  I had seen the decision that leads to color. For PR #124 , I have used info from here. https://github.com/thestr4ng3r/ghidra/blob/6c10f36f06468f866188cccf960c019779fb9028/Ghidra/Features/Decompiler/src/decompile/cpp/variable.cc#L452-L470

In there also, the same logic (if I didn't miss out anything) is used there to classify local variables and parameters. This is the reason why I used color to classify a variable as a parameter and local variable. Have you seen examples where color is mistakenly annotated?

NirmalManoj: Also note that in some cases, the XML has just the color of the variable e.g. ``` <variable color="param">param_1</variable> ```
In other cases where variable has an attribute `varref`, it is first checked to make sure that it's not a global variable or a constant variable.

NirmalManoj: Earlier, forgot to add the following check, I have added it now.
```
	else (!varnode->getHigh()->isPersist() && (varnode->getHigh()->getSymbol() != (Symbol *)0))
		AnnotateLocalVariable(node, out);
```

thestr4ng3r: Do you know the place in `printc.cc` where these cases without `varref` like `<variable color="param">param_1</variable>` are emitted?

NirmalManoj: @thestr4ng3r I'm not sure if I understand the question. If you were asking to confirm if `<variable color="param">param_1</variable>` is given the color="param" from the location you pointed to, then yes it happens at the place [you linked.](https://github.com/thestr4ng3r/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc#L1560-L1563)

thestr4ng3r: Yes, that is what I meant. Now the question is why exactly some variables don't get the `varref` attribute.

NirmalManoj: @thestr4ng3r I had noticed that it doesn't get the `varref` attribute when in places where variables are declared. For reference, let's use [example_pdgx_pdg_output.zip](https://github.com/radareorg/r2ghidra-dec/files/5006086/example_pdgx_pdg_output.zip). In the case of local variable `pvVar2`, it's used in three places.
Declaration.
`<type color="type" id="0x80feab8523c497cf">void</type>
        <syntax></syntax>
        <op>*</op>
        <syntax></syntax>
        <variable color="var">pvVar2</variable>`
Rest
1. `<variable color="var" varref="0x3c4" opref="0x27d">pvVar2</variable>`
2. `<variable color="var" varref="0x3c4" opref="0x1d">pvVar2</variable>`

NirmalManoj: @thestr4ng3r Do you agree that the current code will work in cases where we have `varref`? From what I understand, in cases where we have varref, we are doing exactly what's done in [variable.cc(#L452-L470)](https://github.com/thestr4ng3r/ghidra/blob/6c10f36f06468f866188cccf960c019779fb9028/Ghidra/Features/Decompiler/src/decompile/cpp/variable.cc#L452-L470
) when we have the check
```
	else if (!varnode->getHigh()->isPersist() && (varnode->getHigh()->getSymbol() != (Symbol *)0))
		AnnotateLocalVariable(node, out);
```

thestr4ng3r: Ah I see. Yes it would work but using color is still conceptually wrong. Moreover it currently uses the printed text `node.child_value()` as the variable name which is only marginally better than just "grepping".
What you do get however in the case of the declaration is the following:
```
<vardecl symref="0x400001000000000d">
        <type color="type" id="0x80feab8523c497cf">void</type>
        <syntax></syntax>
        <op>*</op>
        <syntax></syntax>
        <variable color="var">pvVar2</variable>
</vardecl>
```
So you can get the true info about the symbol itself from the parent's `symref`:
```
<symbol name="pvVar2" id="0x400001000000000d" cat="-1">
...
```
which also has the `cat` field indicating whether it is a parameter or regular variable.

NirmalManoj: Thanks a lot, this is really helpful. Here's what I am planning to do now.
Make a Symbol_ID -> Symbol map.

1. If varref exists, get Symbol from the varnode by `varnode->getHigh()->getSymbol()`.
2. Otherwise, get symbol from the map.

In order to get the `symref`, I will do `node.parent()` which will give the `vardecl` node and then get the attribute `symref`.

thestr4ng3r: Yep, that approach sounds good!

NirmalManoj: @thestr4ng3r @karliss I have pushed changes now.

NirmalManoj:
@thestr4ng3r I have added tests.

PR #131, opened 2020-08-15
Add Windows CI

yossizap:
**Detailed description**

Added a new github actions workflow for windows to find issues earlier and share an msys-based build script with Cutter.
I tried to get the build to work with VS2019 but GNUwin32's bison for windows is one major behind the one we use in Linux and seems to freeze forever without any errors while building. Kept msys for now since I was able to control the dependencies.

**Test plan**

Build with windows


yossizap:
Tests require the Sleigh Plugin which will solved in a separate PR, I don't see a simple solution here.

To compile the Cutter plugin we'll need to download cutter-deps for Qt, does that sound reasonable?

XVilka:
I think it's good time to update it, especially with cutter-deps now build with GHA:
- https://github.com/rizinorg/cutter/pull/2490
- https://github.com/rizinorg/cutter-deps-qt/
- https://github.com/rizinorg/cutter-deps

scripts/msys_build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+mkdir -p build && cd build || exit 1
+cmake ../ -G Ninja \
+    -DCMAKE_C_COMPILER=cl \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_PREFIX_PATH="$INSTALL_PREFIX;$INSTALL_PREFIX/include/librz;$INSTALL_PREFIX/include/librz/sdb" \
```
thestr4ng3r, line 7: Is this really necessary (the explicit `/include/...`)? Sounds more like something to be handled by `FindRizin.cmake`

yossizap: We do the same in Cutter but I can try to make it work

https://github.com/rizinorg/cutter/blob/9c7ef2465fd57c6ffa70126d0350b15d0ea1eb41/dist/CMakeLists.txt#L112

thestr4ng3r:
superseded by #211

PR #132, opened 2020-08-24
Implement decompiler isRunning() 

NirmalManoj:
**Detailed description**

Implementing decompiler `isRunning()` was missed out while refactoring the decompiler interface in r2ghidra-dec. This PR fixes it.

...

**Test plan**
1. Check code.
2. Due to some regression as pointed out by @karliss [in this discussion,](https://github.com/radareorg/cutter/pull/2402#discussion_r475795188) this can't be tested right now. You can modify the code to maybe set `task = DecompilerRunning` to make sure the decompiler interface in Cutter able to get the required data from the r2ghidra-dec.

...

**Closing issues**

...


ITAYC0HEN (approved):
good catch guys :)

PR #140, opened 2020-09-02
Add ARMv8, PPC64, SystemZ Travis

XVilka:
**Detailed description**

...

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,44 @@
-
 language: cpp
 
 branches:
   only:
     - master
 
-os: linux
-dist: bionic
+jobs:
+  fast_finish: true
+  include:
+    - os: linux
+      name: X86_64
+      dist: focal
+    - os: linux
+      name: S390X
+      arch: s390x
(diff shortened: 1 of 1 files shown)


XVilka:
Is it possible to build without Cutter counterparts? Because seems there is no Qt for these architectures.

thestr4ng3r:
yes

XVilka:
Ok, now it builds successfully on multiple platforms but many tests are broken.

thestr4ng3r:
Didn't look too closely but the thest are broken because the plugin isn't loaded at all.
Also x86 can be removed completely from travis since it is now tested in GHA.

XVilka:
Yes, I know it should be removed. I kept it as a reference for now, to judge that it should work.

XVilka:
@thestr4ng3r it's ready to be merged now. ARMv8 build passes fine, S390 and PPC are marked as "allowed failures" like in Rizin, to be able to improve the support later.

PR #156, opened 2020-09-22
Fix pdg when using asm.arch=r2ghidra, uses asm.cpu

trufae:
**Detailed description**

pdg takes the arch from asm.arch.. which is fine for capstone plugin for example. but when using the r2ghidra asm/anal plugins, this information must be parsed from asm.cpu

**Test plan**

r2 -a r2ghidra -e asm.cpu x86 -
af
pdg

**Closing issues**

...

### DIFF
--- a/src/ArchMap.cpp
+++ b/src/ArchMap.cpp
@@ -149,7 +149,22 @@ std::string SleighIdFromCore(RCore *core)
 {
 	const char *arch = r_config_get(core->config, "asm.arch");
 	auto arch_it = arch_map.find(arch);
-	if(arch_it == arch_map.end())
-		throw LowlevelError("Could not match asm.arch " + std::string(arch) + " to sleigh arch.");
+	if(arch_it == arch_map.end()) {
+		char *cpu = strdup (r_config_get(core->config, "asm.cpu"));
+		char *colon = cpu;
+		while (*colon) {
+			if (*colon == ':') {
+				*colon = 0;
+				break;
+			}
+			*colon = tolower (*colon);
+			colon++;
+		}
+		arch_it = arch_map.find(cpu);
+		free (cpu);
+		if(arch_it == arch_map.end()) {
+			throw LowlevelError("Could not match asm.arch " + std::string(arch) + " to sleigh arch.");
+		}
+	}
 	return arch_it->second.Map(core);
 }
--- a/src/ArchMap.cpp
+++ b/src/ArchMap.cpp
@@ -158,9 +158,7 @@ std::string SleighIdFromCore(RCore *core)
 	}
 	auto arch_it = arch_map.find(arch);
 	if(arch_it == arch_map.end()) {
-		if(arch_it == arch_map.end()) {
-			throw LowlevelError("Could not match asm.arch " + std::string(arch) + " to sleigh arch.");
-		}
+		throw LowlevelError("Could not match asm.arch " + std::string(arch) + " to sleigh arch.");
 	}
 	return arch_it->second.Map(core);
 }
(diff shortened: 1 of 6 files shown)


src/ArchMap.cpp
```diff
@@ -149,7 +149,22 @@ std::string SleighIdFromCore(RCore *core)
 {
 	const char *arch = r_config_get(core->config, "asm.arch");
 	auto arch_it = arch_map.find(arch);
```
trufae, line 151: maybe if (!strcmp (arch, "r2ghidra")) ?

thestr4ng3r: Yes, better check this right away and use the same sleigh id as the asm/anal plugins instead of just falling back to it.

src/ArchMap.cpp
```diff
@@ -149,7 +149,22 @@ std::string SleighIdFromCore(RCore *core)
...
+		char *colon = cpu;
+		while (*colon) {
+			if (*colon == ':') {
+				*colon = 0;
+				break;
+			}
+			*colon = tolower (*colon);
+			colon++;
+		}
+		arch_it = arch_map.find(cpu);
```
thestr4ng3r, line 163: This is wrong, `cpu` is the sleigh id while `arch_map` is keyed on r2 plugin names, so they are unrelated.

thestr4ng3r (changes requested):

src/ArchMap.cpp
trufae: i know. thats why im splitting the asm.cpu string to the first colon which basically gives me the arch name. and this thing maps fine with the r2 plugin. otherwise its impossible to use the decompiler when using the ghidra disassemblers. which solution do you propose here?

thestr4ng3r: > and this thing maps fine with the r2 plugin.

If it does, it is only by coincidence.

> which solution do you propose here?

Return the `asm.cpu` value as-is. It is already the full sleigh id you want.

trufae:
check again

trufae:
umf seems like this is the wrong place to implement this... do you know where's the line? i added a printf, but i dont get a single message with this oneliner

```
$ r2 -qcq -c"pd 10" -a r2ghidra -e asm.cpu=v850 /bin/ls
            ;-- main, entry0, sym.func.100001060:
            0x100001060      No    sleigh specification for v850
            0x100001061      No    sleigh specification for v850
            0x100001062      No    sleigh specification for v850
            0x100001063      No    sleigh specification for v850
            0x100001064      No    sleigh specification for v850
            0x100001065      No    sleigh specification for v850
            0x100001066      No    sleigh specification for v850
            0x100001067      No    sleigh specification for v850
            0x100001068      No    sleigh specification for v850
            0x100001069      No    sleigh specification for v850
$ r2 -qcq -c"pd 10" -a r2ghidra -e asm.cpu=V850:LE:32:default /bin/ls
       ::   ;-- main, entry0, sym.func.100001060:
       ::   0x100001060      divh  r21,  r9
       ``=< 0x100001062      bnc   0x100001022
            0x100001064      st.b  r10,  0x5641[r1]
            0x100001068      sst.w r10,  0x80[ep]
            0x10000106a      sld.h 0x82[ep], r10
            0x10000106c      divh  r19,  r9
            0x10000106e      sst.h r29,  0x2[ep]
            0x100001070      addi  0x0,  r24,  r0
            0x100001074      and   r9,   r17
            0x100001076      cmp   r23,  r8
```

thestr4ng3r:
You aren't calling pdg at all in you examples.

The change is still wrong. It should be just something like:
```c++
std::string SleighIdFromCore(RCore *core)
{
	const char *arch = r_config_get(core->config, "asm.arch");
	if (!strcmp(arch, "r2ghidra"))
		return r_config_get(core->config, "asm.cpu");
	// ...
```
Please also make sure to keep the correct style (allman etc.)

thestr4ng3r:
Also would be good to have a test for this

trufae:
i'll add tests , you can review now. it's working properly

thestr4ng3r:
Superseded by #164

PR #159, opened 2020-09-24
Use dummy RIO for disassembling when none is available

FXTi:
**Detailed description**

https://github.com/radareorg/r2ghidra/pull/152

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/src/asm_ghidra.cpp
+++ b/src/asm_ghidra.cpp
@@ -11,6 +11,7 @@ static SleighAsm sasm;
 static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len)
 {
 	int r = 0;
+	RIO *rio = nullptr;
 	if(!a->cpu)
 		return r;
 #ifndef DEBUG_EXCEPTIONS
@@ -18,7 +19,14 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len)
 	{
 #endif
 		RBin *bin = a->binb.bin;
-		sasm.init(a->cpu, bin? bin->iob.io: nullptr, SleighAsm::getConfig(a));
+		if(!bin)
+		{
+			rio = r_io_new();
+			r_io_write_at(rio, 0, buf, len);
+			sasm.init(a->cpu, rio, SleighAsm::getConfig(a));
+		}
+		else
+			sasm.init(a->cpu, bin->iob.io, SleighAsm::getConfig(a));
 
 		sasm.check(a->pc, buf, len);
 		r = sasm.disassemble(op, a->pc);
@@ -30,6 +38,8 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len)
 		r = 1;
 	}
 #endif
+	if(rio)
+		r_io_free(rio);
 	op->size = r;
 	return r;
 }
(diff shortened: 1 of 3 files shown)


src/asm_ghidra.cpp
```diff
@@ -11,14 +11,22 @@ static SleighAsm sasm;
...
 		return r;
 #ifndef DEBUG_EXCEPTIONS
 	try
 	{
 #endif
 		RBin *bin = a->binb.bin;
-		sasm.init(a->cpu, bin? bin->iob.io: nullptr, SleighAsm::getConfig(a));
+		if(!bin)
+		{
+			rio = r_io_new();
```
FXTi, line 24: @trufae I think RIO here is wrongly created.

FXTi:
```bash
build [rasm2 *] % rasm2 -a r2ghidra -c x86:LE:32:default:gcc -d 89e1     fx-ti@DELL
JNC 0x67
```

@trufae

src/asm_ghidra.cpp
trufae: i dont see why is this needed

src/asm_ghidra.cpp
```diff
@@ -30,6 +38,8 @@ static int disassemble(RAsm *a, RAsmOp *op, const ut8 *buf, int len)
 		r = 1;
 	}
 #endif
+	if(rio)
+		r_io_free(rio);
```
trufae, line 42: the if (notnull)free(notnull) is a bad nattern

trufae (commented):
uhm can you add a test? not sure about some of those changes

src/asm_ghidra.cpp
FXTi: Change to what?

thestr4ng3r: Just `r_io_free(rio)` without the if. `free(NULL)` in C is valid and r2's free functions follow this pattern too.

FXTi:
Test added.

src/asm_ghidra.cpp
FXTi: Because in SLEIGH, it need RIO, or something like it.

src/asm_ghidra.cpp
FXTi: get it.

trufae:
ci looks red

FXTi:
Absolutely. is my RIO init process wrong? This RIO reads out wrong bytes in my local debugging environment.

thestr4ng3r:
is `a->pc` maybe not 0? Initialization itself should be ok, but I am not sure you can directly write into the RIO like this. You might either need to enable cache or add a writeable map there.

trufae (commented):
Use riobind. Dont created new instances

src/asm_ghidra.cpp
trufae: why? i dont use the instance from bin , instead of creating a dummy instance with no maps loaded or anything setup?  this looks wrong to me. write fill fail below and all the reads will be ffff

src/asm_ghidra.cpp
trufae: you are not the owner of rio in bin.

PR #164, opened 2020-10-09
Fix pdg with sleigh plugin and add short asm.cpu form

thestr4ng3r:


trufae:
Can we have this without depending on pdga?

thestr4ng3r:
Yes

PR #165, opened 2020-10-11
Fix null derefs after r_reg_get

trufae:
**Detailed description**

Segfaults

**Test plan**

```
$ r2 -a r2ghidra -b 32 -e asm.cpu=v850 -s 0xc8d0 -c 'wx 8007 d0c8;ao' -
WARNING: No calling convention defined for this file, analysis may be inaccurate.
zsh: segmentation fault  r2 -a r2ghidra -b 32 -e asm.cpu=v850 -s 0xc8d0 -c 'wx 8007 d0c8;ao' -
$
```

**Closing issues**

none registered

### DIFF
--- a/src/anal_ghidra.cpp
+++ b/src/anal_ghidra.cpp
@@ -2722,8 +2722,7 @@ static void sleigh_reg_set_float(RReg *reg, const char *name, int type, bool F)
 static bool sleigh_reg_get_float(RReg *reg, const char *name, int type)
 {
 	RRegItem *tmp = r_reg_get(reg, name, type);
-	if(tmp)
-		return tmp->is_float;
+	return tmp? tmp->is_float: false;
 }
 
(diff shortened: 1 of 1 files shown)


src/anal_ghidra.cpp
```diff
@@ -2715,13 +2715,15 @@ static bool sleigh_esil_signext(RAnalEsil *esil)
...
+		tmp->is_float = F;
 }
 
 static bool sleigh_reg_get_float(RReg *reg, const char *name, int type)
 {
 	RRegItem *tmp = r_reg_get(reg, name, type);
-	return tmp->is_float;
+	if(tmp)
+		return tmp->is_float;
 }
```
thestr4ng3r, line 2727: missing return

thestr4ng3r (changes requested):

PR #217, opened 2021-03-22
Add noreturn show in function signature

### LINKED ISSUE #86: Show noreturn information in the function signature
by @radare
```
[0x100004434]> pdg

void sym.imp.exit(void)
{
    // WARNING: Could not recover jumptable at 0x000100004434. Too many branches
    // WARNING: Treating indirect jump as call
    (*_reloc.exit)();
    return;
}
[0x100004434]>
```

before (void) will be good to have the 'noreturn' attribute

Yuuoniy:
**Detailed description**
Show noreturn attribute  for no return function
now the expected output like this:
```c
void sym.imp.exit noreturn (void)
{
    // WARNING: Treating indirect jump as call
    (*_reloc.dyld_stub_binder)();
    returndebug in rz_core_annotated_code_print;
}
```
**Test plan:**
seek to a noreturn function and `pdg` to see.

**Closing issue:**
closes #86 

**Testbin**
[rr.zip](https://github.com/radareorg/r2ghidra-dec/files/3670319/rr.zip)

### DIFF
--- a/test/db/extras/ghidra
+++ b/test/db/extras/ghidra
@@ -3117,3 +3117,27 @@ undefined8 main(void)
 }
 EOF
 RUN
+
+
+NAME=show noreturn
+FILE=rizin-testbins/elf/ioli/crackme0x07
+CMDS=<<EOF
+aaa
+s sym.imp.exit
+pdg
+EOF
+EXPECT=<<EOF
+
+// WARNING: Switch with 1 destination removed at 0x080483e8
+// WARNING: [rz-ghidra] Failed to match type signed int for variable var_ch to Decompiler type: Unknown type identifier
+// signed
+// WARNING: [rz-ghidra] Detected overlap for variable var_dh
+
+void sym.imp.exit noreturn (void)
+{
+    // WARNING: Treating indirect jump as call
+    (**(code **)0x8049ffc)(0x30);
+    return;
+}
+EOF
+RUN


thestr4ng3r (changes requested):
Thanks for the pr! This is unfortunately not the right place to change the code output. It should be done in the initial code printing part before the xml even gets created. For this, you will want to look at the `PrintC` class in ghidra: https://github.com/rizinorg/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc
In rz-ghidra we use a subclass already to change its behavior: https://github.com/rizinorg/rz-ghidra/blob/master/src/RizinPrintC.h
If you can add the noreturn just by overriding a function in `RizinPrintC` without duplicating a significant amount of code from ghidra, adding it here would be good, otherwise you can send a pr to https://github.com/rizinorg/ghidra to change it in `PrintC`.

Yuuoniy:
OK, I have taken a look at `PrintC`,  ghidra uses different emit functions to output every part of a function，eg. https://github.com/rizinorg/ghidra/blob/master/Ghidra/Features/Decompiler/src/decompile/cpp/printc.cc#L2289  used to generate FunctionDeclaration.  So my first idea is add a noreturn node when generating XML based on `FuncProto.isNoreturn` attribute.
How about this?

thestr4ng3r:
Yes, this looks good. Now please update the ghidra submodule on rz-ghidra side here and add a test case to https://github.com/rizinorg/rz-ghidra/blob/master/test/db/extras/ghidra, ideally on top of an existing binary from https://github.com/rizinorg/rizin-testbins

Yuuoniy:
I've updated.

thestr4ng3r (approved):
Perfect, thanks!

PR #233, opened 2021-05-07
Fix tests to use `rizin =` instead of `rizin -`

XVilka:
Because after 0.2.[01] version Rizin swapped `-` and `=`, where `-` now reads from stdin.
Thus all `FILE=-` should have been replaced with `FILE==`.

### DIFF
--- a/test/db/extras/analysis_ghidra
+++ b/test/db/extras/analysis_ghidra
@@ -197,7 +197,7 @@ EOF
 RUN
 
 NAME=v850cmp
-FILE=-
+FILE==
 EXPECT=<<EOF
 ghidra
 V850:LE:32:default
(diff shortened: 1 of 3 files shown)


XVilka:
Hmm:
```

[XX] /build/test/db/extras/analysis_ghidra v850cmp
rizin -escr.utf8=0 -escr.color=0 -escr.interactive=0 -N -Qc 'e asm.arch=ghidra
e asm.cpu=v850
e asm.parser=v850.pseudo
e asm.bits=32
pdga
e analysis.arch
e analysis.cpu
wx e731
ao 1
' -
-- stdout

-- stderr
^D

-- exit status: 1
```

But this is unrelated anyway, so I guess PR can be merged anyway.

PR #327, opened 2023-08-09
fix string being directly passed to printf (no "%s")

chayleaf:
hopefully there was no way there would be a `%` in the string either way, but this fixes compiling with some additional errors enabled (maybe `-Wformat -Wformat-security -Werror=format-security` would do it; when I attempted packaging rz-ghidra for NixOS it only errored out on MacOS, so perhaps there's a difference in behavior between gcc and clang)

There's another call to esilprintf with an empty string which causes a warning (`-Wformat-zero-length`).


XVilka (approved):
BTW, just for the record, at some point ESIL will be retired once we finish the migration to RzIL: https://github.com/rizinorg/rizin/issues/2080

PR #377, opened 2026-02-18
Use C++20 for build

kazarmy:
This pr uses C++20 for the build to fix errors of the following form in rizinorg/cutter#3554:

ghidra-c++20

(https://github.com/rizinorg/cutter/actions/runs/22096335762/job/63854269453#step:9:12682)

The reasoning for the change is basically the same as in https://github.com/rizinorg/jsdec/pull/67#issuecomment-3914066220.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,10 +23,10 @@ endif()
 
 if(CMAKE_VERSION VERSION_LESS "3.1")
 	if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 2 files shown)


kazarmy:
The builds compile (in Travis), and the redness appears unrelated (or should be solved by rizinorg/cutter#3554 but there's kind of a cyclic dependency going on here).

kazarmy:
The build also compiles locally under Ubuntu 24.04.

notxvilka (approved):

PR #379, opened 2026-04-04
Fix build due to RzAsm public struct removal

kazarmy:
This pr fixes the build with regard to rizinorg/rizin#6089.


kazarmy:
It builds on Travis (and locally), Travis X86_64 shows a slight regression from 20 to 22 XX compared to dev, but unknown whether it is related to this pr. The Cutter build errors should be fixed by rizinorg/cutter#3600, but they cannot be fixed now because the dependency is cyclic.

notxvilka (approved):

PR #383, opened 2026-04-27
Fix Windows MSVC build failures and harden internal API declarations

ZeroNinx:
## Problem

rz-ghidra did not build cleanly on Windows with MSVC.

The failure showed up as a combination of compile-time and link-time issues, including incorrect Windows code paths, unresolved exported symbols, and MSVC incompatibilities in a few source locations.

## Root cause

There were multiple independent Windows-specific problems:

- the MSVC build path did not explicitly enable UTF-8 or standard C++ exception handling
- some Ghidra code paths expected `_WINDOWS` to be defined
- `rz_ghidra_lib_init` and `rz_ghidra_lib_fini` were not exported for Windows DLL consumers
- `rz_ghidra_internal.h` relied on indirect macro visibility for `RZ_API` and used a non-unique include guard
- the floating-point overflow helpers used `NAN`, which is not reliable in the affected MSVC build path

## Fix

This PR updates the Windows build and source declarations to make the MSVC path explicit and self-contained:

- add MSVC compile options for `/utf-8` and `/EHsc`
- define `_WINDOWS` on Windows builds
- set `CMP0177` to `NEW` when available
- mark `rz_ghidra_lib_init` and `rz_ghidra_lib_fini` with `RZ_API`
- make `rz_ghidra_internal.h` self-contained by giving it its own include guard and including `rz_types.h`
- replace `NAN` with `std::numeric_limits<long double>::quiet_NaN()`

## Validation

Validated on Windows with MSVC and Ninja.

Steps:
- initialized the required recursive submodules
- configured rz-ghidra against an existing Rizin installation prefix
- built the project with MSVC

Result:
- configuration succeeded
- build completed without the previous Windows-specific failures
- only non-fatal upstream warnings remained


thestr4ng3r:
@ZeroNinx What are the exact errors from the compiler that you encountered?

ZeroNinx:
@thestr4ng3r 

I re-tested the issue against a local Windows/MSVC environment.

The relevant fix is commit `112928f` (`Fix MSVC build, handle export symbols, and sanitize internal header`). Its parent is `e52894b`.

Build environment used for reproduction:

- Visual Studio 2022 generator: `Visual Studio 17 2022`
- Platform: `x64`
- Toolset host: `host=x86`
- MSVC: `19.38.33143.0`
- Rizin was built and installed locally, then used through `CMAKE_PREFIX_PATH`
- CMake was pointed at the locally installed `RizinConfig.cmake`

The parent commit `e52894b` fails first when building `analysis_ghidra` with MSVC:

```text
src\analysis_ghidra.cpp(2399,36): error C2065: “NAN”: undeclared identifier
src\analysis_ghidra.cpp(2444,36): error C2065: “NAN”: undeclared identifier
src\analysis_ghidra.cpp(2489,36): error C2065: “NAN”: undeclared identifier
src\analysis_ghidra.cpp(2534,36): error C2065: “NAN”: undeclared identifier
```

This is fixed in `112928f` by replacing the four `NAN` uses with:

```cpp
std::numeric_limits<long double>::quiet_NaN()
```

After applying only that `NAN` replacement locally to the parent commit, the build gets past `analysis_ghidra` and then fails while linking `asm_ghidra.dll`:

```text
asm_ghidra.obj : error LNK2019: unresolved external symbol "void __cdecl rz_ghidra_lib_init(void)" (?rz_ghidra_lib_init@@YAXXZ) referenced in function "bool __cdecl init(void * *)" (?init@@YA_NPEAPEAX@Z)
asm_ghidra.obj : error LNK2019: unresolved external symbol "void __cdecl rz_ghidra_lib_fini(void)" (?rz_ghidra_lib_fini@@YAXXZ) referenced in function "bool __cdecl fini(void *)" (?fini@@YA_NPEAX@Z)
Release\asm_ghidra.dll : fatal error LNK1120: 2 unresolved externals
```

This is fixed in `112928f` by exporting `rz_ghidra_lib_init()` and `rz_ghidra_lib_fini()` with `RZ_API` in both `src/rz_ghidra_internal.h` and `src/core_ghidra.cpp`.

`112928f` also adds the MSVC/CMake setup needed for this environment:

```cmake
if(POLICY CMP0177)
	cmake_policy(SET CMP0177 NEW)
endif()

if(MSVC)
	add_compile_options(
		"$<$<COMPILE_LANGUAGE:C>:/utf-8>"
		"$<$<COMPILE_LANGUAGE:CXX>:/utf-8>"
		"$<$<COMPILE_LANGUAGE:CXX>:/EHsc>"
	)
endif()

if(WIN32)
	add_compile_definitions(_WINDOWS)
endif()
```

It also fixes `src/rz_ghidra_internal.h` by using a correct internal include guard and including `<rz_types.h>`, which is required for `RZ_API`.

With the same Windows/MSVC environment, `112928f` builds successfully, while its parent `e52894b` does not.

PR #387, opened 2026-05-26
Update Ghidra to 12.1 release

thestr4ng3r:
* PrettyXmlEncode removed because ghidra can pretty-print now too
* zlib dependency needed for compressed sleigh
* pugi::parse_ws_pcdata_single used for correct whitespace parsed from pretty-printed xml code
* Some formatting options are now available upstream


notxvilka (changes requested):
```
D:\a\rz-ghidra\rz-ghidra\build\_deps\zlib-src\zlib.h(34): fatal error C1083: Cannot open include file: 'zconf.h': No such file or directory
```

Issue #375, opened 2026-01-26
Imports symbols are not resolved


wargio:
```
sub.msvcrt.dll__XcptFilter();
0x1400018ee      jmp     qword [_XcptFilter] ; 0x140002240
```

```c
void sub.msvcrt.dll__XcptFilter(void)
{
    // WARNING: Treating indirect jump as call
    (*(code *)0x2ab6)();
    return;
}
```

another one, notice the `arg1_00 = (*(code *)0x2bec)(0);`

```c

uint64_t fcn.140001958(int64_t arg1)
{
    uintmax_t arg1_00;
    int64_t iVar1;
    
    arg1_00 = (*(code *)0x2bec)(0);
    if ((arg1_00 != 0) && (iVar1 = fcn.1400018fc(arg1_00), iVar1 != 0)) {
        if (*(int16_t *)(iVar1 + 0x5c) == 2) {
            return 2;
        }
        if (*(int16_t *)(iVar1 + 0x5c) == 3) {
            return 1;
        }
    }
    return arg1 & 0xffffffff;
}
```

```
fcn.140001958(int64_t arg1);
; arg int64_t arg1 @ rcx
0x140001958      push    rbx
0x14000195a      sub     rsp, 0x20
0x14000195e      mov     ebx, ecx  ; arg1
0x140001960      xor     ecx, ecx  ; LPCWSTR lpModuleName
0x140001962      call    qword [GetModuleHandleW] ; 0x1400021c0 ; HMODULE GetModuleHandleW(LPCWSTR lpModuleName)
0x140001968      test    rax, rax
0x14000196b      jz      0x140001995
0x14000196d      mov     rcx, rax  ; uintmax_t arg1
0x140001970      call    fcn.1400018fc ; fcn.1400018fc
0x140001975      test    rax, rax
0x140001978      jz      0x140001995
0x14000197a      mov     ecx, 0x02
0x14000197f      cmp     word [rax+0x5c], cx
0x140001983      jnz     0x140001989
0x140001985      mov     eax, ecx
0x140001987      jmp     0x140001997
0x140001989      cmp     word [rax+0x5c], 0x03
0x14000198e      mov     eax, 0x01
0x140001993      jz      0x140001997
0x140001995      mov     eax, ebx
0x140001997      add     rsp, 0x20
0x14000199b      pop     rbx
0x14000199c      ret
```

[calc.tar.gz](https://github.com/user-attachments/files/24854024/calc.tar.gz)

[...]

Issue #229, opened 2021-04-19
Show Function Names in Decompiled Source Code


TheFrozenDuck:
While playing PlaidCTF2021 I noticed that rz-ghidra does no show the function name in the decompiled code, rather than a call to a function pointer.

In the disassembly the function names are shown
![image](https://user-images.githubusercontent.com/5834688/115215746-16d60d00-a104-11eb-9c62-f4c7eab6256e.png)

and in the decompiled code you can only see the function pointer.
![image](https://user-images.githubusercontent.com/5834688/115215967-543a9a80-a104-11eb-9d22-f846898a2b29.png)

XVilka:
@TheFrozenDuck do you have a small test case to reproduce the issue?

Issue #69, opened 2020-01-14
Remove pointer wrapper on imported function calls


ITAYC0HEN:
Currently r2ghidra-dec shows calls to imported function like this:
```
(*_VirtualProtect)(ppcVar4, 4, 0x40, &lpflOldProtect);
```
But the expected output should look like this:
```
VirtualProtect(ppcVar4, 4, 0x40, &lpflOldProtect);
```

Both Ghidra and IDA treat it like this. Here is an example from GHIDRA:

![image](https://user-images.githubusercontent.com/20182642/72352101-56849c80-36ea-11ea-9cc6-86552db1f399.png)

XVilka:
Please provide a sample binary to test on.

XVilka:
@ITAYC0HEN could you please provide a sample/test?

PR #392, opened 2026-08-01
Do not propagate import slots as read-only constants

fr1pp:
### Problem

On PE, calls to imported functions decompile to a raw pointer instead of the API name:

```c
(*(code *)0x2af4)("case 1");
```

The IAT sits in `.rdata` (MSVC merges `.idata$5` into it), so `getReadonly()` hands it to the decompiler as constant data. A call through a slot then folds into the file's at-rest bytes before `RizinScope` can name it.

### Fix

Subtract the import slots from the read-only ranges, instead of turning read-only propagation off globally with `ghidra.ropropagate=0`. Everything else that is read-only keeps propagating.

Flags in executable mappings are skipped, so ELF is unaffected. There the flags sit on PLT stubs, and the GOT slots behind them already resolve by name through their relocations.

This leaves the `(*name)(...)` call wrapper (#69) in place, which is a separate matter of registering imports as functions rather than code pointers. Its wider cause is that the PE loader attaches no import info to the IAT relocations (#375), which also disables the reloc-based tail-call override in `PcodeFixupPreprocessor`. Fixing that in rizin would cover several PE cases at once.

### Test

`readonlypropagate imports (PE)`, next to the existing `readonlypropagate (objc)` in `test/db/extras/ghidra`, decompiles `main` in `rizin-testbins/jmptbl/test_switch_indirect.exe`. The 14 imported calls gain their names (`(*(code *)0x2af4)` becomes `(*_sym.imp.api_ms_win_crt_stdio_l1_1_0.dll_puts)`) while the 99-case switch table and its strings stay intact.

- fails on `dev`, passes with this change
- `rz-test -L db/extras`: 54 OK / 1 BR / 0 XX, the BR being the pre-existing `NAME=typedef`

Built against rizin `dev` HEAD.

> [!NOTE]
> Opus 5 was used to investigate the behavior and to draft the patch and test, which I reviewed and verified myself.

Ref #375 (the PE raw-pointer half). Related to #69 and #229.

### DIFF
--- a/src/RizinLoadImage.cpp
+++ b/src/RizinLoadImage.cpp
@@ -20,6 +20,40 @@ void RizinLoadImage::loadFill(uint1 *ptr, int4 size, const Address &addr)
 	rz_io_read_at_mapped(core->io, addr.getOffset(), ptr, size);
 }
 
+namespace {
+
+struct ImportSlotCtx {
+	RangeList *list;
+	AddrSpace *space;
+	RzIO *io;
+	ut64 slot_size;
+};
+
+// Punch one import slot out of the read-only ranges. The loader owns a slot's
+// value, so it is a property of a process and never a constant of the program.
+// Unbound it is a hint/name table RVA; in a dump, one run's resolved address.
+bool subtract_import_slot(RzFlagItem *fi, void *user)
+{
+	auto ctx = reinterpret_cast<ImportSlotCtx *>(user);
+	RzIOMap *map = rz_io_map_get(ctx->io, fi->offset);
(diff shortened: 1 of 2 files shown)


thestr4ng3r:
Thanks for the patch, the changes generally look good to me. Please also specify which LLM/AI tools you used for investigation and implementation, and how, if you used any.

fr1pp:
Sorry, it was in the original checklist, which I dropped after realising that isn't the convention here. Re-added the disclosure to the description.

